PR #18, opened 2026-02-17
Add Cargo.lock to MANIFEST.in

jameshilliard:
Some build systems expect packages to include a Cargo.lock file, this ensures builds are reproducible.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,2 +1,2 @@
-include pyproject.toml Cargo.toml
+include pyproject.toml Cargo.toml Cargo.lock
 recursive-include src *


kurok:
Thanks @jameshilliard — shipping `Cargo.lock` for reproducible builds is the right instinct, but it turns out this is already covered, so I'm going to close this.

This project uses the **maturin** build backend (`build-backend = "maturin"`), which generates the sdist from the cargo/git file set and **already includes `Cargo.lock`** by default. Building the current `master` sdist confirms it:

```
$ maturin sdist
$ tar tzf fast_mail_parser-0.3.0.tar.gz | grep Cargo.lock
fast_mail_parser-0.3.0/Cargo.lock
```

The same sdist also contains `.github/`, `Makefile`, `LICENSE`, `tests/`, `rust-toolchain`, etc. — none of which are listed in `MANIFEST.in` — which shows the maturin backend doesn't consult `MANIFEST.in` at all (it's effectively vestigial here; setuptools is never invoked). So adding `Cargo.lock` to that `include` line wouldn't change the produced artifact.

Net: reproducible sdist builds already work today. Closing as already-satisfied — thanks for flagging it!

jameshilliard:
So all that's needed is a new fixed release? The [latest sdist](https://files.pythonhosted.org/packages/dd/b1/af43ef34a7c2560d38ec7949b190744025d3c5f2c4d3e8c657185563e1e3/fast_mail_parser-0.2.5.tar.gz) is missing the Cargo.lock.

kurok:
Yes, but give me one day to deliver more fixes planned.  And solve problems with pypi prev owners. So if I can publish or find owner will do release.  Other wise move to another org.  Will keep you updated.

kurok:
@jameshilliard publish under question till monday I think
so you can use new version from here
command for install 
`pip install "git+https://github.com/namecheap/fast_mail_parser.git@v0.4.0"`

PR #61, opened 2026-06-12
Cache cargo-audit binary to speed up the audit job

kurok:
## Why

The `cargo audit (advisory)` job is the slowest in CI — consistently **~3 min** vs ~30–45s for every other job. `rustsec/audit-check` runs `cargo install cargo-audit`, **compiling cargo-audit from source on every run**.

## What changed

Replace the action with an explicit install whose compiled binary is **cached** (`actions/cache` on `~/.cargo/bin/cargo-audit`):

- **First run / cache-key bump:** `cargo install cargo-audit --locked` (~3 min), then the binary is cached.
- **Subsequent runs:** cache hit → install step skipped → `cargo audit` runs in **seconds**.

The RustSec **advisory DB is still fetched fresh every run** (cargo-audit clones it at runtime), so findings stay current regardless of the cached binary.

## Unchanged behavior

- Still runs `cargo audit` on **stable** (`RUSTUP_TOOLCHAIN` overrides the repo's 1.83 pin — cargo-audit's MSRV is newer).
- Still **advisory** (`continue-on-error`), so a found advisory surfaces in the log without blocking unrelated PRs.

## Note

The first CI run on this PR will still take ~3 min (cache miss, populating the cache). The speedup appears on the next run — I'll trigger a re-run to demonstrate the cache hit and report the before/after timing.

Caches are branch-scoped, so once this merges to `master`, the master-populated cache becomes available to all branches' audit jobs.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -61,23 +61,38 @@ jobs:
   # Supply-chain audit. Advisory (continue-on-error) so a fresh advisory against
   # a pinned dependency surfaces without blocking unrelated PRs — flip to
   # blocking once the dependency stack is current. cargo-audit needs a newer
-  # rustc than the repo's pinned 1.83, so it runs on stable: RUSTUP_TOOLCHAIN
-  # overrides the rust-toolchain file for this step only.
+  # rustc than the repo's pinned 1.83, so the job runs on stable via
+  # RUSTUP_TOOLCHAIN (overrides the rust-toolchain file).
+  #
+  # cargo-audit is `cargo install`ed from source (~3 min — previously the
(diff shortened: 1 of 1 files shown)


kurok:
Measured on this PR (same commit, audit job re-run):

| audit job | duration |
|---|---|
| cache **miss** (first run: `cargo install` + save) | **193 s** (3m13s) |
| cache **hit** (restore binary + scan) | **8 s** |

On the warm-cache run the `Install cargo-audit` step is skipped (`Cache hit for: cargo-audit-bin-Linux-v1`) and `cargo audit` still scans all 22 deps, so findings remain current. ~24× faster on the job, and it removes the ~3-min long pole from CI wall-clock.

Issue #96, opened 2026-07-02
parse_many: parallel batch parsing on a rayon pool (one FFI call, order-preserved, per-item errors)


kurok:
## Problem

The library's whole reason to exist is speed, and #91 already releases the GIL during parsing — but the API is one-message-at-a-time (`parse_email`). High-volume consumers (mail-scanning pipelines, mailbox indexers, spam/abuse processing) parse millions of messages and currently drive parallelism from Python: thread pools juggling individual `parse_email` calls, paying per-call FFI overhead, Python-side scheduling, and result marshaling one object at a time. Rust can do the fan-out natively.

## Proposal

```python
from fast_mail_parser import parse_many

results = parse_many(payloads)                 # list[bytes | str] → list[PyMail | ParseError]
results = parse_many(payloads, threads=8)      # optional cap; default = rayon's default (num CPUs)
```

- One FFI call; the GIL is released once; a rayon pool parses all payloads in parallel; results return **in input order**.
- **Per-item error semantics**: one malformed message must not fail the batch — each slot is either a `PyMail` or a `ParseError` *instance* (not raised), so callers zip inputs to outcomes. (`raise_on_error=True` opt-in for fail-fast callers.)
- DoS limits (#21's input-size and recursion caps) apply per item, unchanged.
- Memory honesty: results materialize all parsed mails at once — documented guidance (chunk your batches; a 10k-message batch of 1 MB mails holds ~everything decoded in RAM). Pairs with the lazy-parsing issue for metadata-only sweeps.

## Acceptance criteria

- [ ] `parse_many` returns order-preserved results for mixed `str`/`bytes` inputs (contract tests mirroring `test_contract.py` patterns for the single-parse API).
- [ ] Per-item errors: batch with 1 malformed of N returns N results with a `ParseError` in the right slot and N−1 parsed mails (test); `raise_on_error=True` raises on first failure.
- [ ] `threads` caps the rayon pool (observable via a large-batch timing test or thread-count instrumentation in a debug hook); default documented.
- [ ] Benchmark added to `tests/benchmark/`: `parse_many` vs Python-side `ThreadPoolExecutor` + `parse_email` over the existing corpus — ≥ 2× speedup on a 4-core runner recorded in the PR and `check_benchmark.py` guards the single-message baseline from regressing.
- [ ] GIL actually released for the batch duration (test pattern from #91: a Python thread increments a counter while a large batch parses; counter progresses).
[...]

PR #169, opened 2026-08-27
Borrow bytes payloads instead of copying them

kurok:
Addresses the remaining piece of #96. **Does not close it** — the issue also covers the rayon pool and API shape, which `parse_many` already landed; whether anything else there is still wanted is worth a separate read.

## The copy

Every payload was copied into Rust-owned memory before parsing began. For `parse_many` that means the whole batch was duplicated *before any of it was parsed*, while the caller still held the originals — ten gigabytes of copies for ten thousand one-megabyte messages. The function's own docstring warned about batch memory without mentioning that half of it was avoidable.

`bytes` payloads are now borrowed via `PyBackedBytes`, which keeps the Python object alive and hands out its buffer. Reading that buffer with the GIL released is sound because `bytes` is immutable — nothing can change or free it while the reference is held.

**`str` still gets copied and cannot not be.** Under the limited API, getting UTF-8 out of a `str` means asking CPython to encode it, which allocates; there is no buffer to borrow. So `bytes` is the fast path, which is also what reading a file or socket hands you. A test asserts both paths give identical results.

## Measuring it honestly took two tricks

I first wrote this test with a batch of garbage payloads and an assertion that they all fail to parse — behaviour I had not verified and cannot run locally, since there's no Rust toolchain in my environment. Replaced with something already proven in `test_dos_limits.py`:

**An oversized payload.** One just past `MAX_INPUT_BYTES` is rejected *before any parsing work*, so the copy is essentially the only memory movement left. For a payload that parses normally, the decoded result dominates and hides the thing under test.

**A subprocess.** `ru_maxrss` is a high-water mark that never falls. `test_dos_limits.py` already allocates ~100 MiB in this process, so `before` would have been sitting above anything the call reaches — and the assertion would have passed while measuring nothing. This is the kind of test that silently succeeds forever, so it's worth being explicit about.

The probe asserts peak growth stays under 20 MiB while rejecting a 100 MiB payload. Reintroducing the copy breaches that by 5x.

## Also

- A single-threaded `parse_many` benchmark: measures per-payload overhead without thread-scheduling noise. Parallel throughput is a separate question.
- Corrects the benchmark file's claim that the gate selects its pair *by substring*. It matches exact names now — which is precisely why a new benchmark can be added there without breaking the gate.

## Note on the gate

This is the first Rust-touching PR since #168, so it's the first to exercise the **interleaved** comparison path rather than the byte-identical skip. I'll check the gate's own output on this PR before merging — the throughput effect may well sit under the noise floor, since the win here is memory rather than time, and I'd rather report that than claim a speedup.

### DIFF
--- a/src/fast_mail_parser.rs
+++ b/src/fast_mail_parser.rs
@@ -18,6 +18,7 @@ mod mail_parser;
 
 use mailparse::MailParseError;
 use pyo3::prelude::*;
+use pyo3::pybacked::PyBackedBytes;
 use pyo3::types::{PyBytes, PyDateTime, PyDict, PyList, PyString, PyTzInfo};
 use pyo3::{create_exception, exceptions, wrap_pyfunction};
 use std::num::NonZeroUsize;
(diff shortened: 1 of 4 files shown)


kurok:
## The gate's verdict — and I was wrong about the size of this

I predicted the throughput effect would sit under the noise floor and that the real win was memory. The measurement says otherwise:

| Benchmark | base | this revision | Delta | |
|---|---|---|---|---|
| `parse_many` (8 × 0.75 MiB, threads=1) | 19.196 ms | 14.035 ms | **−26.9%** | |
| `parse_message` (single) | 1.761 ms | 1.759 ms | −0.1% | |
| `full_read` (single) | 1.785 ms | 1.762 ms | −1.3% | |
| `mail_parser` | 12.516 ms | 12.446 ms | −0.6% | control |
| `mailparser_lib` | 14.418 ms | 13.534 ms | −6.1% | control |
| `stdlib_email` | 18.468 ms | 18.031 ms | −2.4% | control |

Noise floor from the controls: **6.1%**. The batch delta is more than four times that.

### Two arithmetic checks, because a 27% win deserves suspicion

**Does the magnitude make sense?** The old path copied 8 × 0.75 MiB = 6 MiB. The time removed is 5.16 ms, i.e. ~1.16 GiB/s effective. That is the right order for *allocate-then-copy* (each payload got a fresh `Vec`), not for a bare `memcpy` — which is exactly what the old code did.

**Where did the time go?** After the change, the batch takes **14.035 ms**, and eight single-message parses take **8 × 1.759 = 14.07 ms**. Batch time is now pure parsing: the per-payload overhead is not merely reduced, it is gone within measurement error.

The single-message benchmarks barely move, as expected — one 0.75 MiB copy against a 1.76 ms parse is a rounding error. The copy only mattered where the batch made it repeat, which is precisely #96's regime.

### Note on the gate

This is the first PR to exercise the interleaved comparison path added in #168 (the extension hashes differ, so both sides were measured). It also demonstrates the point of that change: at a 6.1% control floor, the old single-shot method could plausibly have reported anything from −20% to −33% for this, or hidden it.

Issue #157, opened 2026-08-27
headers key order is nondeterministic across parses of identical bytes


kurok:
Found by the fuzz harness in #156 on its very first run, from the existing seed corpus.

## What happens

`PyMail.headers` is built from a Rust `std::collections::HashMap`, whose iteration order is randomised per instance. That order becomes the insertion order of the Python dict, and Python dicts preserve insertion order — so the key order differs on every parse of the same bytes.

```python
raw = open("tests/data/rfc/empty_body.eml", "rb").read()
orders = {tuple(parse_email(raw).headers.keys()) for _ in range(40)}
len(orders)          # 40  -- forty parses, forty different orders
```

Content is unaffected, which is why nothing noticed:

```python
parse_email(raw).headers == parse_email(raw).headers   # True
list(parse_email(raw).headers) == list(parse_email(raw).headers)   # False
```

On the 29-header `valid_message.eml`: 20 parses, 20 distinct orders.

## Why it matters

Dict lookup is unaffected, so most consumers never see it. These do:

- `list(mail.headers)` / `for name, values in mail.headers.items()` — iteration order varies run to run.
- `json.dumps(mail.headers)` — different bytes each time, so caches keyed on it miss, and diffs between two runs are noise.
- Any snapshot or golden-file test over the header dict is inherently flaky.
- **The original header order is not recoverable at all.** For forensic or delivery-path work — the use case that motivated #23 — the order of `Received` relative to other trace headers is information the message carried and this API discards.

There is also a documentation mismatch. The 0.7.0 changelog says of #23's fix: *"Each key now maps to every value it appeared with, in message order."* That is true of the values within a key; a reader would reasonably assume the keys are ordered too. They are not, and never match the wire:

```
order on the wire: From, To, Date, Message-ID, Subject, Content-Type, ...
one parse gives:   To, Content-Transfer-Encoding, Date, From, Message-ID, ...
```

## Possible fixes

1. **Preserve wire order** — replace the `HashMap` with an insertion-ordered map. `indexmap` is the usual choice, though it is a new dependency and so touches the licence allowlist and the lockfile. Alternatively a `Vec<(String, Vec<String>)>` converted to a dict at the boundary, which needs no dependency: the PyO3 layer already builds the dict, and Python preserves whatever order it is given.
2. **Sort the keys** — deterministic and dependency-free, but throws away the wire order, which is the more useful of the two properties.

[...]

Issue #150, opened 2026-08-26
Malformed message: body lost and a MIME boundary reported as a header key


kurok:
Found while extending `parse_many`'s test coverage (#149). `tests/data/invalid_message.eml` — a 103 KB real-world Google-delivered message — parses without error, but the result is wrong in two ways, and the stdlib gets it right.

## What happens

```python
raw = open("tests/data/invalid_message.eml", "rb").read()

mail = parse_email(raw)
len(mail.text_plain)        # 0   <- the body is gone
len(mail.headers)           # 31

"--_----------=_MCPart_1735325173" in mail.headers   # True  <- a MIME boundary
```

Compared with the stdlib (`policy=default`) on the same bytes:

| | fast_mail_parser | stdlib `email` |
| --- | --- | --- |
| `text/plain` body parts | **0** | 1 |
| header keys | 31 | 29 |
| MIME boundary present as a header key | **yes** | no |

So a boundary delimiter line is being consumed as a header field, and the body that follows it is lost rather than surfaced. The stdlib recovers both.

## Why it matters

Two things, in order of importance:

1. **It fails silently.** No `ParseError`, no empty-ish result that looks suspicious — just a message whose body is missing and whose header map contains one entry that is not a header. A consumer scanning bodies would treat this message as having none. For a spam classifier or a forensic tool that is a wrong answer, not a missing one.
2. **The name misled two of my own changes today.** The file is called `invalid_message.eml`, so I twice assumed it fails to parse: once in a comment on the parity suite, once in a `parse_many` test that consequently failed. Both corrected in #149. Whatever is decided here, the fixture deserves a name or a comment describing what it actually is.

## Where it likely sits

The message is malformed — that is presumably why it was kept — so the question is not "parse it correctly" but "fail or repair predictably". This is squarely the case #100's warnings channel exists for: parse best-effort, but *record* that a repair happened, so `warnings != []` can route a message to review instead of silently classifying it on a body that was dropped.

Worth deciding between:

- **Warn and keep going** (#100 item 2): emit something like `invalid-header-skipped` / `body-lost`, so the loss is observable.
- **Reject** as a `HeaderParseError`, on the grounds that a boundary appearing where a header is expected means header parsing has genuinely gone off the rails.
- **Repair**, matching the stdlib's recovery. Most work, and it is upstream behaviour in `mailparse` rather than something this crate decides.

## Reproducing

```bash
python - <<'EOF'
import email, email.policy
[...]

Issue #99, opened 2026-07-02
parse_email_tree: full MIME part-tree API with nested message/rfc822 support and stdlib-parity walk()


kurok:
## Problem

`PyMail` is a flattened projection — bodies here, attachments there — and the open classification bugs (#22 phantom parts, #25 ignored Content-Disposition) are symptoms of a deeper limitation: **any** flattening policy is wrong for someone. Forensics tooling needs the exact MIME topology; nested `message/rfc822` (bounced/forwarded mail — ubiquitous in abuse pipelines) is currently smashed into the parent's attachment list with its structure lost; `multipart/alternative` vs `multipart/mixed` distinctions (which HTML body corresponds to which plain-text sibling) are unrecoverable. Python's stdlib `email` has `walk()`; the fast library should not force a fallback to the slow one for structure.

### `parse_email_tree` — the structural API (additive; `parse_email` unchanged)

```python
node = parse_email_tree(payload)     # PyMimePart

# PyMimePart:
#   content_type: str            # "multipart/alternative", "text/plain", ...
#   headers: ...                 # this part's headers (same header semantics as PyMail post-#23)
#   disposition / filename / content_id
#   children: list[PyMimePart]   # empty for leaves
#   content: bytes | None        # decoded leaf content (None for multipart containers)
#   is_message: bool             # message/rfc822 — children contain the parsed inner message's root
```

- `walk()` generator (depth-first, stdlib-compatible order) implemented Python-side over `children` for familiarity.
- **Nested messages parsed, not opaque**: `message/rfc822` parts expose their inner structure as a subtree (bounded by the existing recursion caps from #21 — nesting counts against the same depth limit).
- `PyMail` becomes (or is documented as) a *view* over this tree; the flattening rules that #22/#25 fix get stated in terms of the tree ("a leaf is an attachment iff …") — turning the classifier from folklore into a spec.
- Lazy-mode interaction: `parse_email_tree(payload, mode="metadata"|"lazy")` honors the same modes as the lazy-parsing issue (content deferral per leaf).

## Acceptance criteria

- [x] Tree structure matches stdlib `email.message_from_bytes(...).walk()` topology (content types in order) for the full test corpus + RFC corpus (parity test — the strongest correctness oracle available).
- [x] `message/rfc822` fixtures (single and doubly-nested bounce) expose inner subject/headers/bodies via the subtree; recursion cap fires at the documented depth with a clear `ParseError` (extends `test_dos_limits.py`).
[...]

Issue #97, opened 2026-07-02
Lazy parsing modes: metadata-only sweeps and deferred attachment decoding (mode="metadata"|"lazy")


kurok:
## Problem

`parse_email` always materializes everything: all text bodies decoded, every attachment's full content copied into Python `bytes`. But two of the highest-volume real workloads don't want the payloads:

- **Scanning/triage**: classify by headers + subject + sender across millions of messages — attachment bytes are pure wasted allocation (often 95%+ of the message).
- **Selective extraction**: find the one PDF in a mailbox — decode only that attachment, not every inline image in every message.

For a library whose selling point is throughput, "decode everything always" leaves the biggest single optimization on the table.

### 1. Headers-only mode

```python
mail = parse_email(payload, mode="metadata")
# subject/date/headers fully populated;
# text_plain/text_html: [] ; attachments: metadata only (mimetype, filename, size) with content=None
```

- MIME structure is still walked (part inventory + sizes — cheap); transfer-decoding and content copying are skipped entirely.
- `PyAttachment` gains `size: int` (decoded-size estimate or encoded size, documented which) in both modes — useful generally.

### 2. Deferred attachment content (full mode stays default)

- `parse_email(payload, mode="lazy")`: bodies decoded as today; `PyAttachment.content` decodes **on first access** (the raw part bytes are retained internally; decode happens at property access, cached after). The mailbox-triage sweet spot: full text search + selective attachment pull.
- Default `mode="full"` remains byte-identical to today — zero behavior change without opt-in.

### Design notes

- Sizes and structure honor the existing DoS caps (#21); `metadata` mode makes the caps *cheaper* to enforce, not looser.
- Interaction with the open classification fixes (#22/#25): modes change *when* content is materialized, never *what* is classified — implemented against the current classifier, unaffected by its fix.

## Acceptance criteria

- [x] `mode="metadata"`: headers/subject/date identical to full mode on the whole test corpus (parametrized equality test); bodies empty; attachment metadata (mimetype, filename, size) present with `content is None`.
- [x] `mode="lazy"`: `content` access returns bytes identical to full mode for every attachment in the corpus (equality test); repeated access returns the same object (cache test); un-accessed attachments never decode (instrumented via a decode counter in a test build or a timing proxy — mechanism chosen in review).
- [x] Default mode regression: `parse_email(payload)` output byte-identical to today across the corpus and RFC corpus (`test_rfc_corpus.py`).
[...]

PR #180, opened 2026-08-27
Add parse_email_tree: the MIME structure, not a projection

kurok:
Implements #99's structural API. **Pure addition** — `parse_email` is untouched, so nothing breaks and no release window is needed (unlike #97, where the proposed API widens `content` to `bytes | None`; see [my note there](https://github.com/namecheap/fast_mail_parser/issues/97#issuecomment-5436367860)).

## What it adds

```python
from fast_mail_parser import parse_email_tree, walk

root = parse_email_tree(payload)
for part in walk(root):            # depth first, stdlib walk() order
    print(part.content_type)
```

Nodes carry `content_type`, `headers`, `filename`, `content_id`, `disposition`, `children`, `is_message`, and `content` — the transfer-decoded bytes of a leaf, `None` for a container whose body is only its children with boundaries between them.

## The two things a flat projection cannot say

**Which body goes with which.** A `multipart/alternative` node's children are the plain and HTML renderings *of the same thing*. Through `parse_email` they are one entry in `text_plain` and one in `text_html`, with nothing relating them.

**What is inside a bounce.** `message/rfc822` is now parsed rather than opaque — mailparse hands it over as a leaf, and embedded messages are what bounce and abuse handling is made of. `is_message` is `True` and the embedded message's own root is the part's single child, so its headers are reachable instead of being an attachment blob to re-parse by hand. Doubly-nested bounces work, and that nesting counts against the **same** depth cap as a multipart nest, so an onion of forwards cannot recurse further than a multipart tree can.

## Correctness oracle

Topology is asserted against `email.message.Message.walk()` across the **whole fixture and RFC corpus**, not chosen examples. `invalid_message.eml` is excluded for the reason the parity suite excludes it — the two parsers legitimately disagree about its structure (#150).

Cross-API consistency is also checked: every attachment's bytes from `parse_email` must appear as some leaf's content in the tree.

## Refactor carried along

Header collection is now one shared helper used by both the flat view and the tree, so they cannot drift on what a message's headers are — and the reasoning about key order (#157) lives in one place instead of two.

## Three tests I rewrote before committing

Since I have no Rust toolchain locally and CI is the only thing that can run these, it's worth saying which assumptions I caught myself making:

- asserting random bytes fail to parse — they don't, as #150 established the hard way; now using the two payloads `test_error_taxonomy.py` proves do raise
- excluding `message/rfc822` from the cross-API comparison, where it *does* count as an attachment to the flat API
- a "walk is lazy" test that would have passed just as well against a materialised list; now asserts it is a generator

## Not in this PR

`mode="metadata"|"lazy"` on the tree, per #97 — that decision is open on #97 and lazy decoding needs interior mutability, a different risk and a different review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **`parse_email_tree(payload)`** returns the message's MIME tree with the
+  structure intact, and **`walk(part)`** iterates it depth-first in the same order
+  as the stdlib's `email.message.Message.walk` (#99). A pure addition:
+  `parse_email` is unchanged.
+  - Two things the flat projection cannot express. A `multipart/alternative`
+    node's children are the plain and HTML renderings *of the same thing*, which
+    `text_plain`/`text_html` cannot relate. And a `message/rfc822` part -- a
+    bounce or forward -- is now parsed rather than opaque: `is_message` is `True`
+    and the embedded message's own root is the part's single child, so its
+    headers are reachable instead of being an attachment blob to re-parse.
+  - Embedded nesting counts against the same recursion cap as multipart nesting.
+  - Tree topology is asserted against the stdlib's `walk()` over the whole
+    fixture and RFC corpus, which is the strongest correctness oracle available.
+
 - An internal panic now raises `ParseError` instead of PyO3's `PanicException`
   (#102). This is not about memory safety -- PyO3 already catches panics at the
   boundary, so one never aborted the process -- it is about which `except` clause
--- a/Readme.md
+++ b/Readme.md
@@ -289,6 +289,55 @@ The break-even point is not portable — the parse scales with cores while the
 per-call overhead does not — so treat the second table as the shape of the
 trade-off and measure your own mix if it matters.
 
+### The MIME tree
+
+`parse_email` hands back a flat projection: bodies in one list, attachments in
+another, `multipart/*` containers dropped. That is what most code wants, and it
+throws away the shape of the message. `parse_email_tree` keeps it.
+
+```python
+from fast_mail_parser import parse_email_tree, walk
+
+root = parse_email_tree(payload)
+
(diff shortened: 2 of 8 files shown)


kurok:
## The gate caught a 24% regression in code that never runs, and here is where it was

The first push measured **+29.5%** on every `fast_mail_parser` benchmark against a 2.1% control floor, per-round spread 0.3% — so, real. The tree API is additive and `parse_email` does not call any of it, which made it worth bisecting rather than guessing.

| branch | change | delta |
|---|---|---|
| #181 | header refactor only | +3.3% … +5.0% |
| #182 | refactor + **core** tree code, no binding | +3.4% … +4.9% |
| this PR, first push | everything | **+29.5%** |
| this PR, now | everything + inline hints | **+0.1% … +0.4%** |

So the core half cost nothing beyond the refactor, and ~24 points sat in the **binding layer** — `PyMimePart`, its `#[pymethods]`, the recursive `from_part`, the new pyfunction. None of which executes during `parse_email`.

### Cause

`catch_panics` is generic, so every entry point instantiates it. With a third instantiation, the one wrapping `parse_email` stopped being inlined — and an opaque call with unwind edges around the parse body costs real optimisation *inside* that body. That predicts the effect scaling with message size, which is exactly the pattern: large-message benchmarks +29% to +55%, the small-message one +1.7%.

`#[inline(always)]` on `catch_panics`, plus `#[inline(never)]` on the tree's cold binding functions, removes all of it — including the refactor's few percent.

The comment on `catch_panics` says the attribute is load-bearing, because it looks exactly like decoration someone would tidy away.

### Ruled out along the way

`lto = true` and `codegen-units = 1` are **already** in the release profile. I had a commit written adding them before I checked — worth recording, since a codegen-unit boundary is the natural first guess and would have been a confident no-op.

### What this says about the gate

#168 made this gate interleaved a few hours ago. The old single-shot version had a ~16% artifact against a 7% threshold, so a +29% *might* have shown as anything. It would also have had no control column, and the control is what made this measurement believable enough to spend four runs bisecting. The gate paid for itself on the second real PR through it.

PR #181, opened 2026-08-27
EXPERIMENT (do not merge): isolate the +29% in #180

kurok:
Diagnostic only — **not for merge**.

#180 measured **+29.5%** on every fast_mail_parser benchmark against a 2.1% control floor, per-round spread 0.3%, so it is real. Its diff has two independent parts:

1. extracting header collection into a helper shared by the flat view and the tree
2. adding the tree API (a new recursive function in the same module)

This branch carries **only (1)**. If the gate is clean, the cost is in (2) — most likely code layout, which would be consistent with this crate's known codegen sensitivity (a rustc minor version alone moves it 15–30%). If the gate regresses here, a three-line refactor costs 29% and that is worth knowing on its own.

Closing either way once it reports.

### DIFF
--- a/src/mail_parser.rs
+++ b/src/mail_parser.rs
@@ -191,6 +191,42 @@ fn disposition_token(part: &ParsedMail<'_>, kind: &DispositionType) -> Option<St
     })
 }
 
+/// Every value of every header, keyed by name, in first-appearance key order.
+///
+/// Repeated keys keep all their values: collapsing to one kept only the last,
+/// discarding all but the final `Received`, `DKIM-Signature`, `Received-SPF`,
+/// ... -- which made delivery-path tracing and signature verification impossible
+/// (#12, #23).
+///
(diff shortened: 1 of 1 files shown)


kurok:
Answered, closing. The header refactor alone costs **+3.3% to +5.0%** — small, but above the 1.9% control floor, so real: extracting the loop into a function stops it being inlined into `Mail::new`.

That leaves ~24% of #180's +29.5% attributable to the tree code itself. The size-scaling pattern in #180 (large-message benchmarks +29% to +55%, small-message ones +1.7%) points at `get_body_raw` — the transfer decode, which is the size-scaling work. Giving it a second call site plausibly stopped rustc inlining it into the flat path.

Next: try `lto = "fat"` with `codegen-units = 1` so the compiler sees everything at once, rather than hand-tuning inline attributes against a heuristic.

PR #182, opened 2026-08-27
EXPERIMENT (do not merge): is the +24% in the core or the binding?

kurok:
Diagnostic only — **not for merge**. Follows #181.

- #180 (refactor + core tree + binding + API): **+29.5%**
- #181 (refactor only): **+3.3% to +5.0%**
- this branch (refactor + core tree, no binding, no Python API): **?**

If this regresses ~29%, the cost is in the core — most plausibly `get_body_raw` losing inlining once it has a second call site, which fits the size-scaling pattern (large-message benchmarks +29% to +55%, small-message +1.7%).

If it comes back at ~5%, the cost is in the binding layer instead, and `PyMimePart` plus its `#[pymethods]` block is doing something surprising.

Worth noting what is already ruled out: `lto = true` and `codegen-units = 1` are **already** in the release profile, so this is not a codegen-unit boundary and not something more LTO can fix.

### DIFF
--- a/src/mail_parser.rs
+++ b/src/mail_parser.rs
@@ -191,6 +191,126 @@ fn disposition_token(part: &ParsedMail<'_>, kind: &DispositionType) -> Option<St
     })
 }
 
+/// Every value of every header, keyed by name, in first-appearance key order.
+///
+/// Repeated keys keep all their values: collapsing to one kept only the last,
+/// discarding all but the final `Received`, `DKIM-Signature`, `Received-SPF`,
+/// ... -- which made delivery-path tracing and signature verification impossible
+/// (#12, #23).
+///
+/// A `HashMap` cannot carry the key order: its iteration order is randomised per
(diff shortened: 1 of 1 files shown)


kurok:
Answered, closing. Core tree code alone: **+3.4% to +4.9%** — indistinguishable from the header refactor's +3.3% to +5.0% (#181). So the core half is not where the cost is.

That puts the remaining ~24% in the **binding layer**: `PyMimePart` with its `#[pymethods]`, the recursive `from_part`, and the `parse_email_tree` pyfunction. None of it runs during `parse_email`, so this is a cold path degrading a hot one.

Leading hypothesis: `catch_panics` is generic, and a third instantiation stops the one wrapping `parse_email` from being inlined — which turns the parse body into an opaque call behind unwind edges and costs real optimisation inside it. That would explain why the effect scales with message size.

Trying `#[inline(always)]` on `catch_panics` and `#[inline(never)]` on the tree's cold binding functions next.

Issue #104, opened 2026-07-02
Roadmap: fast_mail_parser 2.0 — batch throughput, structured API, fuzz-trusted, install-anywhere


kurok:
## Where the 10x is

The audit wave (#19–#55, mostly closed) turned fast_mail_parser from a stale binding into a hardened, well-tested, properly-released library. The next multiplier is *capability*: the library parses one message at a time into a minimal flat projection, with strings where structure should be, and its speed advantage stops at the single-call boundary.

1. **Throughput is the product — finish the story.** The GIL is released (#91), but there's no batch API and no way to skip decoding payloads you don't need. `parse_many` + lazy modes turn "fast per message" into "fast per mailbox" — the difference is 5–10× in real pipelines.
2. **Structure, not strings.** Senders/recipients, dates, content-IDs, and MIME topology are all parsed internally by mailparse and discarded at the FFI boundary. Exposing them removes the error-prone re-parsing every consumer does in Python.
3. **Trust for untrusted input.** This library eats attacker-controlled bytes for a living; fuzzing + a panic-containment policy is the missing piece of the hardening story.
4. **Adoption friction.** abi3 wheels end the recurring "new Python, no wheel" breakage (#14/#15's whole history); a stdlib-parity suite + migration guide makes switching a decision, not a research project.

## Dedup note — the open correctness backlog is Phase 0 — ✅ **COMPLETE**

**All closed as of 2026-08-26.** These were part of this roadmap but already filed — they are referenced, not duplicated: #22 (phantom attachment parts), #23 (duplicate header collapse), #25 (Content-Disposition classification), #28 (lossy subject/date), #55 (double decode), #12 (all headers — likely resolved by #23's fix, verify then close), #14/#15 (wheel availability — superseded-by-design via abi3, close from that PR). Several carry `breaking-change`: they should batch into one API-v2 release window together with #98's additions.

## Phase 1 — Throughput (the core promise)

[...]

PR #185, opened 2026-08-27
Add mode="metadata": headers and inventory, decoding nothing

kurok:
Implements the metadata half of #97. Lazy mode is not here — it needs interior mutability with a decode-once cache, a different risk and a different review.

## Built as option C, and why

I [laid out three options on #97](https://github.com/namecheap/fast_mail_parser/issues/97#issuecomment-5436367860) because the API as written carries a breaking change. `attachment.content = None` in metadata mode means the stub's `content` widens to `bytes | None`, which breaks every `mypy --strict` caller of the **default** path for a feature they are not using — and #104 says breaking changes batch into one window rather than trickle.

Overloads on the mode literal give the requested API at no such cost:

```python
parse_email(payload)                    # -> PyMail
parse_email(payload, mode="metadata")   # -> PyMailMetadata
```

`PyAttachment.content` stays `bytes`. Nothing changes for existing callers, typed or not.

## Two departures from the issue's text

**No `text_plain`/`text_html` — absent, not empty.** The issue asked for empty lists. An empty list cannot be told apart from *"this message has no text part"*, so a triage sweep counting bodyless messages would count every message. That is the same class of silent wrong answer as #150, in a feature built for triage sweeps. A missing attribute fails loudly instead, and `parse_email_tree` is there when structure is what you want.

**`encoded_size`, not `size`, and only in metadata mode.** A bare `size` reads as the decoded size, which this mode cannot know without doing the decode it exists to skip — base64 inflates by about a third. It is deliberately *not* added to full mode's `PyAttachment`: that would cost the hot path a header parse per part for a number the caller already has as `len(content)`, in an issue whose whole point is not doing unnecessary work.

## Hot path protected by construction

Today's lesson from #180 was that cold binding code cost the hot path 24% through lost inlining alone. So:

- the core's `Mail` and `Attachment` are **byte-identical** — metadata is a separate entry point, not a mode threaded through `Mail::new`, so no branch lands in the hot loop for the cold path's benefit
- the binding's metadata path is `#[inline(never)]`

The gate will say whether that was enough.

## Tests

Mostly equality against full mode over the whole corpus, because the contract that matters is that a sweep and a full parse never disagree about the same message: envelope, `date_parsed`, headers, all five address fields, and the attachment inventory (count, mimetype, filename, content_id, disposition) are asserted identical per fixture.

Plus: `encoded_size >= len(content)` for every attachment (no transfer encoding shrinks its input, so this holds without knowing which was used), a concrete check that a base64 part reports more wire bytes than decoded bytes, bodies absent rather than empty, described attachments carrying no `content`, `mode="full"` matching the default, an unknown mode raising `ValueError`, and `mode` being keyword-only.

One test pins a real difference rather than an equality: **metadata mode cannot report `DecodeError`**, because it never decodes. A part with a broken transfer encoding passes silently here and raises in full mode. Header errors are reported in both.

## Also

`get_body_encoded` returns `Body` directly and `get_raw` lives on the *variant payloads*, not on the enum — so `body.get_raw()` does not compile. Checked against mailparse's source rather than guessed, which is one CI round trip saved.

A benchmark is included; I'll post the metadata-vs-full ratio from this PR's own gate run against #97's ≥5x target.


kurok:
## Gate: clean, after two attempts

| | base | this revision | delta |
|---|---|---|---|
| `parse_message` | 1.898 ms | 1.902 ms | +0.2% |
| `full_read` | 1.906 ms | 1.908 ms | +0.1% |
| `parse_many` | 15.132 ms | 15.095 ms | −0.2% |
| `parse_tree` | 1.880 ms | 1.886 ms | +0.3% |

Noise floor 0.6%. The first attempt measured **+30.4%**, from a three-arm `match` whose error arm calls `format!`, sitting inside the closure `catch_panics` inlines into `parse_email`. `catch_panics` was already `inline(always)` from #180's fix, so it was not the wrapper this time — it was the closure body. The default path is now one string comparison, with the other modes behind `#[inline(never)]` and the error builder `#[cold]`.

Second time in one day that code which never executes cost this hot path ~30%.

## Metadata mode: 3.69x, not the 5x the issue asks for

| | |
|---|---|
| full parse | 1.902 ms |
| `mode="metadata"` | **0.515 ms** |
| speedup | **3.69x** |

Per-round spread 0.001 ms, so the number is not in doubt. And the fixture is genuinely what the criterion asks for — `large_message.eml` is **99% attachment by decoded content**: 211 KiB PDF + 342 KiB PNG against 7 KiB of body text. So the premise held and the target was simply optimistic.

The ceiling is not our decoding, it is **mailparse's structural parse**. Whatever the mode, `parse_mail` scans all 767 KiB to find the boundaries and parses every header; metadata mode's remaining 0.515 ms is essentially that, plus a part walk and one cheap `get_body_encoded` per part. Skipping the base64 decode of 553 KiB and its copy into Python is worth 73% of the total, and no mode can reclaim the rest without lazy *structural* parsing, which #97 puts out of scope.

I would amend the criterion to the measured figure rather than chase it, the same way #96's 2x was amended on evidence. Recorded on the issue.

Issue #100, opened 2026-07-02
Error taxonomy + parse warnings: typed ParseError subtypes, lossy-repair observability, strict mode


kurok:
## Problem

Error reporting is binary: `ParseError` with a message string, or success. Real mail is messier than that — #24 (fixed) showed decode failures being silently swallowed, and the fix chose between "error" and "best effort" globally. Pipeline consumers need the middle ground: *parse what you can, tell me what was lossy*. Today a message with one undecodable body part either fails entirely or succeeds with silently-mangled content, and the caller can't distinguish pristine output from patched-up output — which matters enormously when the consumer is a spam classifier or forensic tool making decisions on the content.

### 1. Error taxonomy (additive; `ParseError` stays the base)

```python
ParseError                    # base (unchanged — existing except clauses keep working)
├── HeaderParseError          # unrecoverable header-section failure
├── MimeStructureError        # malformed multipart structure / recursion cap (#21) exceeded
└── DecodeError               # body/attachment transfer-decode failure in strict mode
```

- Each carries structured context: `part_path` (e.g. `"1.2"` tree coordinates, aligning with the part-tree issue), `charset`/`encoding` where relevant — machine-usable, not just prose.

### 2. Parse warnings — the lossy-success channel

```python
mail = parse_email(payload)
mail.warnings     # list[ParseWarning]: kind, part_path, detail
# kinds: "charset-fallback", "transfer-decode-lossy", "invalid-header-skipped",
#        "address-unparseable", "date-unparseable", ...
```

- Best-effort behavior (post-#24 semantics) becomes *observable*: every lossy repair the parser performs is recorded. Empty list = pristine parse — a guarantee consumers can finally act on (route `warnings != []` mail to quarantine/manual review).
- `parse_email(payload, strict=True)`: any condition that would emit a warning raises the corresponding typed error instead (validation-pipeline mode).

## Acceptance criteria

- [x] Exception hierarchy exported with stubs; `except ParseError` catches all subtypes (compat test); each subtype raised from a dedicated malformed fixture with `part_path`/context populated (table-driven, extending `test_decode_errors.py` / `test_dos_limits.py`).
- [x] Warning kinds emitted for each documented lossy path: charset fallback fixture, broken quoted-printable, bad address header, bad date — each yields exactly the expected (kind, part_path) and *correct best-effort content* alongside (both asserted).
[...]

PR #186, opened 2026-08-27
Repair a missing header/body separator instead of eating the body (#150)

### LINKED ISSUE #150: Malformed message: body lost and a MIME boundary reported as a header key

kurok:
Closes #150.

## The defect

RFC 5322 ends the header block with an empty line. `tests/data/invalid_message.eml`
— a real Mailchimp-delivered message, bare-LF throughout — never sends one: a
folded ` hello` line is followed directly by the first MIME boundary.

mailparse stops parsing headers only at an empty line, and `parse_header` accepts
a colonless line as a field name with an empty value, so it kept consuming the
body as headers — 73 lines, 31 keys, one of them `--<the boundary the message
declares>`, plus a second `Content-Type` belonging to the first part rather than
to the message. With the opening delimiter eaten, the first part's body sat
before the *next* boundary, which makes it multipart preamble: discarded by
definition. The `text/html` alternative was delimited normally and survived, so
the message came back looking populated with its plain-text body silently gone.

That silence is the argument for repairing rather than only reporting: a parse
that quietly returns less than the message contains is worse than one that fails.

## The repair

The rule is the stdlib's, which names the defect
`MissingHeaderBodySeparatorDefect`: a non-continuation line in the header block
that cannot be a header field ends the header block, and the body starts there.

mailparse owns the header parsing, so the rule is applied to the payload mailparse
is handed rather than to mailparse: `repair_missing_separator` in
`src/mail_parser.rs` scans the header block and, if it finds such a line, returns
a copy of the payload with a single `\n` inserted in front of it. mailparse then
reaches the stdlib's conclusion on its own, with no change to it and no second
header parser in this crate. Where the payload is already well-formed the function
returns `None` and the original bytes are parsed, untouched.

Applied by `parse_email`, `parse_many` and `parse_email_tree`, and to an embedded
`message/rfc822` that the tree parses, so the flat and structural views cannot
disagree about such a message. **Not** applied to a multipart *part*'s headers:
those are parsed inside mailparse's boundary split, which is not reachable from
here, so a part that omits its own separator is still read the way mailparse reads
it. That case would need upstream work.

### Why "cannot be a header field" is narrowed to "contains no colon"

The stdlib's `headerRE` is stricter: a field name may hold only printable ASCII
other than colon, so `Subject : x`, or an 8-bit byte in a field name, ends the
block for it too. mailparse accepts both of those as headers, and demoting them to
body text would trade one silent loss (a body) for another (the rest of the
headers). The colon is the part of the rule that identifies *this* defect. Two
exemptions are the stdlib's own: a line starting with space or tab is a folded
continuation, and a leading `From ` line is an mbox envelope header.

Nothing else changes behaviour for a message that has its separator, and the
narrower rule can only fire on strictly fewer inputs than the stdlib's.

## The exclusions are gone

`test_stdlib_parity.py`, `test_mime_tree.py` and `test_parse_many.py` each skipped
this fixture because the two parsers disagreed about it. All three now compare it
like any other fixture — which is the real validation of the fix, since it puts
the stdlib in charge of judging the result rather than numbers written into a
test.

What makes that safe to assert without a build: **repaired, `invalid_message.eml`
differs from `valid_message.eml` — the same Mailchimp message with its header
block correctly terminated — by exactly one line**, the injected ` hello` fold,
which both parsers read identically as `MIME-Version: 1.0 hello`. `valid_message`
passes the parity suite today with exactly three declared divergences, so
`invalid_message` inherits that profile, and the three entries added to
`DIVERGENCES` are its existing ones:

- `text_plain` / `text_html` — CRLF in the decoded bodies. The file is bare-LF on
  the wire, but its quoted-printable parts spell their line endings `=0D=0A`, so
  the decoded bodies carry CRLF while the stdlib normalises to LF. Divergence 4 in
  `docs/compatibility.md`.
- `headers` — folding whitespace collapsed to one space (divergence 5), plus the
  stdlib re-spelling `Date: Wed,  1 Jul 2020 …` with the day zero-padded
  (divergence 2).

`docs/compatibility.md` is updated for all three, since the parity test fails on a
divergence that is declared but not documented and on one that is documented but
no longer observed.

## Performance

Cost on a well-formed message is one pass over the header block, which ends at the
empty line it has: one comparison per line, nothing read past the separator, no
allocation. Only a defective message allocates, and it allocates once. The
100 MiB input cap is still checked *before* the scan, so an oversized payload is
still rejected without being read or copied (`test_payload_memory.py` measures
exactly that).

The change lives in `src/mail_parser.rs`, the PyO3-free core; no hot function body
changed apart from the two-line dispatch, and the scan carries `#[inline(never)]`
so it cannot grow the parse path's code size — which has cost this crate 24%
before (see the Performance section of `CONTRIBUTING.md`).

## Tests

- `tests/test_known_issues.py` — the four `#150` pins now assert the recovery
  instead of the loss: both parts present, no boundary as a header key, one
  `Content-Type`, and the header set and the recovered plain body compared
  against the stdlib. Three added tests pin the *rule* rather than the fixture: a
  terminated header block is untouched, a colonless line ends the block whether
  or not it is a MIME boundary, and a folded continuation does not.
- The second defect is pinned as unchanged: ` hello` still folds into
  `MIME-Version`, as it does for the stdlib.
- No warnings channel here — that is #100, being worked on in parallel. Happy to
  rebase.

## What I could not verify locally

There is no Rust toolchain in the environment I worked in, so **CI is the first
compile**. Everything above about mailparse's behaviour is read out of
`mailparse` 0.16.1's source rather than assumed:

- `parse_headers` breaks only on `\n` / `\r\n` and errors on a lone CR at a header
  start (so a line beginning with CR is deliberately left alone).
- `parse_header` keeps a colonless line as a key with an empty value.
- `find_from_u8_line_prefix` accepts a boundary delimiter at `ix == ix_start`,
  which is what makes a body that begins immediately with `--boundary` split
  correctly after the repair — the one thing the whole approach rests on.
- `MailParseError` borrows nothing, so parsing from a local buffer is sound.

The behavioural predictions were checked by applying a line-for-line Python port
of the scan to every fixture (it fires on `invalid_message.eml` and on nothing
else), to every message-shaped literal in the test suite (nothing), and by
parsing the repaired bytes with an older installed build to confirm the recovered
part, the 29-key header set and the byte-diff against `valid_message.eml`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,29 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- A message whose header block is never terminated by a blank line no longer
+  loses its first MIME part (#150). RFC 5322 ends the header block with an empty
+  line; real mail sometimes omits it, and the stdlib names the defect
+  `MissingHeaderBodySeparatorDefect`. Without that line the underlying
+  `mailparse` keeps consuming the body as headers -- it stops only at a blank line
+  and accepts a colonless line as a field name -- so on the message this was found
+  on it swallowed the first MIME boundary, which left the first part's body before
+  the *next* boundary: multipart preamble, discarded by definition. The
+  `text/html` alternative was delimited normally and survived, so the message came
+  back looking populated with its plain-text body silently gone, its boundary
+  delimiter reported as a header key, and the first part's own headers merged into
+  the message's (two `Content-Type` values).
+  - The payload is now normalised before it reaches `mailparse`, by the stdlib's
+    rule: a non-continuation line in the header block that cannot be a header
+    field ends the header block, and the body starts there. `parse_email`,
+    `parse_many` and `parse_email_tree` all apply it, so the flat and structural
+    views cannot disagree about such a message.
+  - A message that has its separator -- every other fixture in the corpus, and
+    every well-formed message -- is parsed from the original bytes, unchanged.
+  - `tests/data/invalid_message.eml` is consequently no longer excluded from the
+    stdlib-parity, MIME-tree and `parse_many` corpora. The two parsers now agree
+    on its topology, its header set (29 keys, not 31) and both of its bodies; what
+    is left is the CRLF and unfolding divergence that every real message shows.
 - `headers` keys are now in the order the header names first appeared in the
   message, stably across parses. They came from a Rust `HashMap`, whose iteration
   order is randomised per instance, and that order became the Python dict's
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -24,6 +24,12 @@ library reported every MIME node as an attachment and collapsed repeated headers
 to their last value, so the structural dimensions diverged on nearly every
 multipart message.
 
+It also includes `invalid_message`, a real message whose header block is never
+terminated by a blank line. It was excluded from the comparison until this
+library learned the stdlib's recovery for that defect (#150); the two now agree
+on its structure, its header set and its bodies, and the differences left are the
+line-ending and unfolding ones below.
+
 ## Where they differ
 
 ### 1. Raw UTF-8 in headers — this library is more correct
@@ -45,14 +51,16 @@ If you are migrating *to* the stdlib, this one will cost you.
 
 | | |
 | --- | --- |
-| Fixture | `attachment_message` |
+| Fixtures | `attachment_message`, `invalid_message` |
 | Dimension | `headers` |
 | stdlib | `Date: Wed, 24 Apr 2019 10:05:02 +0200` |
 | this library | `Date: Wed, 24 Apr 2019 10:05:02 +0200 (CEST)` |
 
 `headers` gives values as they appeared, unfolded and RFC 2047-decoded, and
 nothing more. The stdlib additionally parses structured headers and re-emits a
-canonical form, dropping the RFC 5322 comment `(CEST)`.
+canonical form, dropping the RFC 5322 comment `(CEST)`. It re-spells them too:
+`invalid_message` sends `Date: Wed,  1 Jul 2020 05:33:42 +0000` and the stdlib
+returns the day zero-padded.
 
 Use `date_parsed` when you want an interpreted value; use `headers` when you want
 what the sender actually wrote.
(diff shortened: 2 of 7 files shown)


kurok:
Closing the loop on the "CI is the first compile" caveat in the description: all 14 checks are green on the first run.

- **Test matrix** — py3.11/3.12/3.13/3.14 pass with `invalid_message.eml` compared like any other fixture in all three corpora. The predicted divergence profile held exactly: the three `DIVERGENCES` entries are observed and none is stale, so the parity suite accepts the message on the header set, the topology and both bodies.
- **Benchmark quality gate** — worst treatment delta **+0.4%** (`test__fast_mail_parser___full_read`), verdict *no significant difference* against the 7% threshold. The extra pass over the header block does not show up, and the codegen lottery did not bite.
- **Lint** (`cargo fmt --check`, mypy, ruff), **fuzz short pass**, **cargo audit/deny**, **CodeQL** — all pass.

Issue #102, opened 2026-07-02
Fuzzing: cargo-fuzz targets, CI jobs, and a panic-to-ParseError policy at the FFI boundary


kurok:
## Problem

This library's job description is "parse attacker-controlled bytes in Rust with unsafe FFI boundaries at both ends" — mail pipelines feed it whatever the internet sends. The audit hardened known cases (#19 lossy conversion, #21 DoS caps, #24 decode errors) and built a strong fixture corpus (`tests/generate_rfc_corpus.py`, `test_dos_limits.py`), but fixtures only re-test known failure shapes. There is no fuzzing: nothing systematically searches for the *unknown* panic, OOM, or pathological-complexity input — and a panic across the PyO3 boundary aborts the host process, so in a mail pipeline a single crafted message can take down the worker.

### 1. cargo-fuzz targets

- `fuzz/` with libFuzzer targets: `fuzz_parse_email` (raw bytes → full parse), `fuzz_parse_metadata` (once lazy modes land), each asserting the invariants that must survive arbitrary input: **no panic**, DoS caps respected (bounded output sizes / recursion), and parse-twice determinism (same input → same output).
- Seed corpus from the existing test fixtures + RFC corpus generator (the investment in #33/#35 becomes fuzz seeds for free); minimized corpus committed to the repo.

### 2. CI integration

- **PR job**: short deterministic fuzz pass (e.g. 60s per target, fixed seed) — catches gross regressions without flaky CI.
- **Scheduled job**: weekly longer run (30+ min per target) on the default branch; new crashers uploaded as artifacts and auto-filed/refreshed into a pinned issue (the drift-alarm pattern).
- Any crasher found gets minimized and committed as a regression fixture in `tests/` — fuzz findings permanently join the corpus.

### 3. Panic policy at the FFI boundary

- Audit: every Rust `panic!`/`unwrap`/slice-index reachable from parse paths either becomes a `ParseError` (coordinating with the error-taxonomy issue) or is proven unreachable; `catch_unwind` at the PyO3 boundary as the last-resort backstop converting surviving panics into `ParseError` instead of process abort (with a debug log of the panic payload).

## Acceptance criteria

- [ ] Both fuzz targets build and run locally via documented `cargo +nightly fuzz run` instructions (CONTRIBUTING section); seed corpus wired from existing fixtures.
[...]

PR #187, opened 2026-08-27
Fuzz the newer APIs for agreement with the flat parse

kurok:
Toward #102, which asked for a metadata fuzz target once the mode existed. It exists now (#185), and the tree (#99) had none either.

## Agreement, not just absence of panics

Both new APIs are **re-derivations of the same message**. A re-derivation's interesting bugs are not panics — they are the cases where it quietly disagrees with the original. So the new `parse_agreement` target asserts:

- **metadata mode succeeds wherever the full parse does.** It does strictly less work: it never transfer-decodes, and `get_body_encoded` cannot fail. The converse is deliberately *not* asserted — a broken transfer encoding fails the full parse and passes metadata, which is documented behaviour, not a bug.
- **the envelope is identical** — subject, date, and the whole header map. A triage sweep and a full parse disagreeing about the same message is exactly what this exists to prevent.
- **the attachment inventory matches** field by field (mimetype, filename, content id, disposition); only content and size differ by design.
- **`encoded_size` is never below the decoded size**, since no transfer encoding shrinks its input.
- **the tree parses deterministically**, and every attachment the flat parse reports appears as some leaf's content — containment rather than equality, since the tree keeps parts the projection drops.

These are the same invariants the corpus tests assert across nineteen fixtures, generalised to arbitrary input. The corpus tests can only tell you the fixtures agree.

## Runner cost held flat

The deep run becomes a **matrix over targets**, so each keeps its own corpus cache — and the key preserves the existing `v2-parse_email` lineage rather than resetting the 2,615 inputs accumulated so far.

The per-target budget is **halved** (1800s → 900s), so two targets cost the same weekly runner time as one did, rather than double. Actions minutes here are a stated concern, and silently doubling a scheduled job is not the way to spend them. Same for the PR job: 30s per target, one minute total, unchanged.

## Notes

`cargo fmt --all` covers workspace members and the fuzz crate is deliberately standalone, so the new target is not format-gated — but it is compiled and run for 30s by this PR's own fuzz job, which is the real check on whether these invariants hold.

### DIFF
--- a/.github/workflows/deep-fuzz.yml
+++ b/.github/workflows/deep-fuzz.yml
@@ -6,6 +6,11 @@ name: Deep fuzz
 #
 # Two things make a weekly run worth more than 30x the PR job:
 #
+# Two targets run in parallel, each with its own corpus: `parse_email` on the flat
+# parse, and `parse_agreement` checking the metadata and tree APIs against it
+# (#97, #99). The per-target budget is halved so the weekly runner cost is
+# unchanged.
+#
 #   - The corpus is cached and restored, so coverage compounds week over week
 #     instead of restarting from the fixtures every time. This is most of the
 #     value: a fuzzer's reach is mostly a function of the corpus it inherits.
@@ -32,7 +37,7 @@ on:
     inputs:
       seconds:
         description: "Fuzzing time per target, seconds"
-        default: "1800"
+        default: "900"
       inject_canary:
         description: "Plant the canary input to test the crash-reporting path"
         type: boolean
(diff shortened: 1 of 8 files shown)


kurok:
## The target earned itself in one run

Rebasing onto #186 made the fuzz job fail immediately, on the fixture that motivated all of this. The two header maps for `invalid_message.eml`:

- **full parse** — one `Content-Type`, no boundary-shaped key
- **metadata mode** — two `Content-Type` values, a `--_----------=_MCPart_...` key, and a stray `Content-Transfer-Encoding`

The full parse was repairing the missing header/body separator; metadata mode was not. `mode="metadata"` landed while #186 was being written, so it was the one entry point the repair never reached.

**Nothing else would have caught it.** `test_metadata_mode.py` still excluded `invalid_message.eml` — the only fixture with the defect — so the exclusion that existed *because* the parsers disagreed was hiding a fresh disagreement. Fourteen green checks on #186 and nineteen fixtures said nothing.

Both fixed here: the entry point repairs, and the exclusion is gone so the corpus test covers it too.

## Then the gate caught me

The first fix inlined the repair into `parse_email_metadata`'s body and cost **+28% on the flat path and +95% on metadata mode** — from a scan that reads **1,729 bytes of a 767 KB message** and fires on none of the well-formed fixtures. I checked that locally rather than assuming: the scan stops after 32 header lines, 0.2% of the file. So it was codegen, not work.

#186 split `Mail::new` into a wrapper over an unchanged body and measured +0.4%. Same shape applied here, and the same result:

| | base | this revision | delta |
|---|---|---|---|
| `parse_message` | 1.898 ms | 1.898 ms | −0.0% |
| `parse_metadata` | 0.516 ms | 0.516 ms | +0.1% |
| `parse_tree` | 1.877 ms | 1.886 ms | +0.5% |
| `full_read` | 1.903 ms | 1.910 ms | +0.3% |

Noise floor 0.4%. The repair costs metadata mode nothing measurable.

Third time today that a few lines near a parse function cost 25–95% while doing almost no work, and the third time the same remedy fixed it. That pattern is now in CONTRIBUTING.md.

Issue #101, opened 2026-07-02
abi3 stable-ABI wheels + free-threaded CPython support (end the new-Python-no-wheel cycle)


kurok:
## Problem

Wheels are built per-Python-version, which is why distribution keeps breaking: #15 (2022-era release uninstallable), #14 (no wheels past 3.10) — every new CPython release requires a publish-matrix change and a release, and users on the newest Python hit `pip` source builds (needing a Rust toolchain) or hard failures. PyO3 supports the **stable ABI (abi3)**: one wheel per platform covers every CPython ≥ the baseline, permanently ending the "new Python, no wheel" class of issue. Separately, free-threaded CPython (3.13t/3.14t) is where a GIL-releasing Rust parser shines brightest — and it needs explicit support declaration.

### 1. abi3 wheels

- Enable `pyo3/abi3-py311` (matching `requires-python >= 3.11`): the publish matrix collapses from {platform × Python-version} to {platform} — 3–4 wheels total instead of a dozen-plus.
- Keep per-version **test** matrix (abi3 changes what we ship, not what we test): tests still run on 3.11–3.14.
- Measure before committing: abi3 disallows some fast-path PyO3 APIs; run the benchmark suite abi3 vs current — if regression > 5%, document the decision either way in the issue (the escape hatch: ship abi3 for breadth + version-specific wheels for the latest CPython, maturin supports both).

### 2. Free-threaded CPython

- Audit the extension for free-threaded safety (PyO3 ≥ 0.23 supports declaring it; the lazy-decode caching from the lazy-parsing issue is the main shared-state risk — coordinate) and publish `cp313t`/`cp314t` wheels once green.
- The README benchmark story gets its best chapter: `parse_many`-style parallelism on free-threaded Python with zero GIL contention (benchmark numbers included).

## Acceptance criteria

- [x] abi3 wheel built for each supported platform installs and passes the full test suite on 3.11, 3.12, 3.13, and 3.14 (CI matrix: build once, test everywhere — the abi3 promise verified, not assumed).
- [x] Benchmark delta abi3 vs version-specific measured and recorded; decision (full abi3 / hybrid) documented in the PR per the >5% rule; `check_benchmark.py` thresholds adjusted deliberately if needed.
- [x] Publish workflow matrix reduced accordingly; OIDC Trusted Publishing + attestations (#26/#89) preserved; a new CPython minor releasing requires **zero** repo changes for installability (stated as the design goal; verified logically in review).
[...]

Issue #21, opened 2026-06-12
Harden against untrusted-input DoS: cap input size and MIME recursion depth


kurok:
**Finding (High).** `extract_mail_parts` recurses per subpart with no depth limit (`mail_parser.rs:78–89`), and every part's raw body is materialized into a `Vec<u8>` then copied again into Python `bytes` (`mail_parser.rs:53–57`, `fast_mail_parser.rs:25`) with no input-size cap. A deeply nested `multipart/*` message can overflow the stack (process abort); thousands of parts or large base64 parts exhaust memory. The repo's own commit `e55a519` ("replace malicious large_message.eml") shows prior awareness.

**Fix.** Enforce limits at the entry point, before/while parsing:
- Input size cap (~25MB) → raise `ValueError` (or `ParseError`) when exceeded
- Recursion-depth cap (~20) → return partial result instead of overflowing

**Acceptance**
- [ ] Over-cap input raises a catchable exception (not a crash)
- [ ] A 25-level-nested MIME fixture parses without stack overflow
- [ ] Limits are documented

_Audit ref: M1-T3. Caps are placeholders — see Open Question on threat model._

PR #198, opened 2026-08-27
Decode attachment content on first access, with mode="lazy" (#97)

kurok:
Closes the open half of #97: `mode="metadata"` landed in #185, `mode="lazy"` did not. This is it.

## The API

```python
parse_email(payload)                    # -> PyMail          (unchanged)
parse_email(payload, mode="full")       # -> PyMail
parse_email(payload, mode="lazy")       # -> PyLazyMail      new
parse_email(payload, mode="metadata")   # -> PyMailMetadata
```

`PyLazyMail` is `PyMail` with one difference: `attachments` are `PyLazyAttachment`.

```python
mail = parse_email(payload, mode="lazy")
mail.subject, mail.text_plain, mail.warnings      # identical to full mode

pdf = next(a for a in mail.attachments if a.mimetype == "application/pdf")
pdf.encoded_size, pdf.is_decoded                  # (215_411, False) -- choose without decoding
data = pdf.content                                # decoded here, and only this one
assert pdf.content is data                        # every later read is the same object
```

`PyLazyAttachment` is `PyAttachment`'s fields plus `encoded_size` and `is_decoded`, with `content` a property that decodes on first access and caches. The mode picks the return type through the stub's `@overload`s on the mode literal, so no caller of the default path sees a change — `PyAttachment.content` is still `bytes` and still costs what it always cost.

**A new type rather than a lazier `PyAttachment.content`.** Changing what an existing attribute costs, and where it raises, is a change to a shipped contract, and #104 batches those into one API-v2 window. Adding a type is neither breaking nor in need of a window — the same reasoning that gave metadata mode its own attachment type instead of widening `content` to `bytes | None`.

## Why the deferred decode is the same bytes

It retains each part exactly as it sits in the message and re-parses that copy on access. That is mailparse's `raw_bytes`, which for a subpart is `&raw_data[ix_part_start..ix_part_end]` — the part's own headers, the separator and its still-encoded body, with the boundary lines excluded and the trailing CRLF stripped by `strip_trailing_crlf`. Both the part's header slice and its body slice are taken from inside that same slice, and `get_body_raw` depends only on the `Content-Transfer-Encoding` header and the body bytes (verified in `body.rs`: `Body::new` uses `ctype` only for charset, which `get_body_raw` never applies). So the re-parse yields the same encoding and the same body, and the decode is byte-for-byte the full parse's `content`.

That is a claim about mailparse's slicing rather than about this crate, so it is tested as one, three ways:

- **the corpus** — equal for every attachment across every fixture and the whole RFC corpus (`tests/test_lazy_mode.py`);
- **the fuzz target** — `parse_agreement` now decodes every retained part and asserts equality with the full parse, plus agreement on the envelope, both body lists and the whole warning list;
- **a local randomized sweep** while writing it — 27,097 mutated fixtures that the full parse accepted, 4,517 attachments compared, zero mismatches and zero cases where the deferred decode failed where the parse had not.

## Numbers

`tests/data/large_message.eml`: 767 KiB, 99% attachment by decoded content (a 211 KiB PDF and a 342 KiB PNG against 7 KiB of text). **From the gate's own run on this PR** — median of three interleaved rounds, each value a benchmark's minimum:

| | | vs full |
| --- | --- | --- |
| `mode="metadata"` | 0.516 ms | 3.87x |
| `mode="lazy"`, nothing read | 0.562 ms | **3.55x** |
| `mode="full"` | 1.995 ms | 1.00x |
| `mode="lazy"`, every attachment read | 2.038 ms | 0.98x |

So an untouched lazy parse lands within 9% of metadata mode — which decodes *nothing*, not even the bodies — and reading everything through it costs about 2% over the default. Deferring saves ~70% when you were not going to decode everything and ~nothing when you were, and the README says plainly to use the default mode in the second case.

Two benchmarks are added (`parse_lazy_untouched`, `parse_lazy_all_attachments`), both guarded with `pytest.skip` so the base side of the gate stays green. The guard catches `ValueError` as well as `TypeError`: the base already has `mode=`, so it rejects an unknown mode rather than rejecting the argument — a `TypeError`-only guard, as the metadata benchmark has, would have failed the base run.

## Thread safety

`content` is a `OnceLock<Py<PyBytes>>` on the Python object, and the GIL is released for the decode so concurrent readers overlap rather than serialise.

Two threads can arrive at an undecoded part together and both decode. That is deliberate: `OnceLock` has no fallible `get_or_try_init` on stable, and initialising through an infallible closure would mean either panicking on a broken transfer encoding or caching the failure. Duplicate work under a race is the cheaper defect, it is bounded, and the property callers actually depend on is unaffected — the cell is published once and every caller, winner or loser, returns the object it holds. A failed decode is not cached at all.

`tests/test_lazy_mode.py` has 16 threads on a `Barrier` hammering one attachment 20 times each: all 320 reads must be the same object and none may fail, plus a second test with 8 threads across all attachments.

**The free-threading audit (#101) survives, which its own note doubted** — it named a decode cache as the obvious way to break "no shared mutable state in the parsing core". The cache is in the binding layer, on the object that owns the bytes; the core's contribution is `decode_part`, a pure function of `&[u8]`. The module note is updated to say so rather than left pointing at a hazard that was avoided.

## Acceptance criteria

- [x] `content` equals full mode's bytes for every attachment across the fixture and RFC corpus
- [x] repeated access returns the same object — and the attachment list hands back the same parts, without which the cache would be per-read and meaningless
- [x] un-accessed attachments never decode. Asserted, not timed: `is_decoded` is public, so "reading one attachment leaves the others untouched" is a test. It is also a real answer for a caller holding an inventory — whether reading `content` is free or is about to cost a decode. The timing proxy is in the table above as well.
- [x] thread safety, with the test described above
- [x] default mode unchanged; stubs, README, CHANGELOG, and `cargo fmt` / clippy / ruff / `mypy --strict` all clean. The overloads are checked rather than assumed: `reveal_type` under `mypy --strict` gives `PyMail` / `PyMail` / `PyLazyMail` / `PyMailMetadata` for the four spellings, and `mode="metadata", strict=True` is rejected statically.
- [x] the #21 DoS caps hold in this mode too (input size and MIME depth), pinned in `tests/test_lazy_mode.py`
- [x] `strict=True` decided and stated — see below

## `strict=True` with `mode="lazy"`: accepted, and it means the same thing

Lazy mode decodes every body part exactly as full mode does, and the one attachment-level repair the parse can report — a quoted-printable escape passed through as literal text (#194) — is found by scanning the *encoded* bytes, which this mode still does. So `warnings` is the same list, asserted per fixture across the corpus and on arbitrary input by the fuzz target. `strict=True` therefore rejects exactly what it rejects in full mode. Metadata mode still refuses it, for the reason it always did, and the error message now names both modes that work.

## The honest costs

**A `DecodeError` moves from the parse to the attribute.** A part whose transfer encoding is broken fails `parse_email` in full mode; here the message parses and only that attachment raises. Usually the more useful behaviour, and pinned as a test rather than left to be discovered — like metadata mode's inability to report one at all.

**Memory for decoding.** The encoded bytes of every attachment are retained until the message is dropped, and base64 is ~1.33x the size of what it encodes, so a retained part costs *more* than the decoded bytes it avoids producing. Right for one PDF out of twenty attachments; wrong for all twenty. The bytes are copied rather than borrowed, which is required: `bytes` payloads are borrowed since #96, so a retained view into the caller's buffer would be a use-after-free waiting for a collection. `test__attachments_outlive_the_payload` pins that.

## On the inlining cliff

`parse_email`'s body is byte-for-byte what it was. The new mode is answered inside the `#[inline(never)]` helper that already handled metadata mode, and **without adding a `catch_panics` call site** — the two modes share one, because that generic instantiates per call site and #99 is the record of a third instantiation costing the hot path 24% while the new code never executed. Strict mode likewise gained a branch inside its existing `catch_panics` rather than a second one. Everything else new is behind `#[inline(never)]`, with `#[cold]` on the decode and on the error builders.

**The gate passed** against current master: `parse_message` **+1.5%**, `parse_metadata` **-0.0%**, `parse_tree` **-1.3%**, worst treatment delta **+1.5%**, under a 1.1% pure-Python noise floor. Verdict: *no significant difference*. An interleaved local A/B on arm64 macOS against the same base agrees (+0.5% worst).

One earlier run of this branch *did* fail the gate at +15.2% on every decoding benchmark while metadata mode moved +0.1%, and it is worth recording why it was not this change: the branch was based on the commit before 0.8.0 was cut, so the PR's merge ref carried master's version bump while the fixed comparison base did not. The gate was measuring a 0.8.0-versioned head against a 0.7.0-versioned base — different `-Cmetadata`, different symbol hashes, different code layout — which is precisely the ±15-26% alignment sensitivity `rust-toolchain.toml` and CONTRIBUTING.md document. Rebasing onto master put both sides on 0.8.0 and the difference disappeared — and the gate's own `.so` hashes prove the binary was never the variable: the head build that measured **+15.2%** against the 0.7.0 base is `66dc7783...`, and the head build that measured **+0.2%** against the 0.8.0 base is `66dc7783...`. The same bytes. Worth knowing for the next branch that sits open across a release: **a stale base can make this gate blame the PR for the release commit's codegen.**

## Not verified / noticed

- The **free-threaded (3.13t/3.14t) build is not exercised by CI**, so the thread tests run under the GIL, where the release-during-decode window is real but the borrow rules are not the free-threaded ones. The reasoning for soundness there is in the `decode` doc comment; a `--disable-gil` job is a separate change.
- The Readme's strict-mode section still ends with "there is no `transfer-decode-lossy` kind" while the table three paragraphs above it lists one — #194 added the kind and left that sentence. Untouched here to keep this diff about lazy mode; worth a one-line fix.
- `Cargo.lock` carried `version = "0.7.0"` for this package after 0.8.0 was cut, so any build dirtied the tree. Synced in the last commit, since a rebase across the release surfaced it.


kurok:
Verified independently before merging: rebased onto 0.8.0, CHANGELOG entry under the empty `## [Unreleased]`, README documents `mode="lazy"` (which the drift test in #196 now asserts is an accepted mode), and the final gate reads **+0.5% worst against a 1.6% noise floor**.

Two things in this PR are better than what I asked for:

**`is_decoded` as a public attribute** turns "an attachment nobody reads is never decoded" from a timing argument into an assertion — and it answers a real question for a caller holding an inventory, namely whether reading `content` is free or is about to cost a decode. I had suggested a timing proxy; this is strictly better.

**The stale-base diagnosis.** The earlier +15.2% was not this change, and the `.so` hashes prove it rather than arguing it: the head build that measured +15.2% and the one that measured +0.2% are byte-identical. The gate had compared a 0.8.0 merge ref against a 0.7.0 fixed base.

That is a flaw in the gate, not in this PR, and I am fixing it separately — it resolves the base from `pull_request.base.sha`, which is the base branch as of the PR's last event and can lag the merge ref. It also means I merged the release cut (#197) reading only "no failures": that run in fact measured **+3.8% to +4.4%** on every decoding benchmark, above its noise floor and under the threshold. Recorded on the follow-up.

PR #203, opened 2026-08-27
PROBE (do not merge): what did the version bump cost?

kurok:
Diagnostic only — **draft, not for merge** (merging would un-release 0.8.0).

#197 cut the release. Its only Rust-affecting content was `Cargo.toml`'s version string, and its gate measured **+3.8% to +4.4%** on every decoding benchmark — above its noise floor, under the 7% threshold — and I merged it having read only "no failures".

This branch reverts **only** the version, against current master. Now that the gate resolves its base as `origin/master` (#200), the delta it reports is exactly what that string is worth:

- **near −4%** → the release cut really did cost that, and 0.8.0 as published is a few percent slower than 0.7.0 by codegen accident. Worth documenting, and worth knowing this crate is that sensitive to a metadata change.
- **near 0%** → #197's run was noise sitting above its own floor, and the published release is fine. Also worth knowing, because I have been treating that gate's numbers as authoritative all day.

Either way the answer goes on the issue and the branch is closed.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,7 +3,7 @@ name = "fast_mail_parser"
 description = "Very fast Python library for .eml files parsing."
 edition = "2021"
 rust-version = "1.83"
-version = "0.8.0"
+version = "0.7.0"
 authors = ["Andrii Sokyrko <wartwvister@gmail.com>"]
 readme = "Readme.md"
 # Both fields: `license` is the SPDX expression tooling reads (cargo-deny warns


kurok:
Answered. Reverting **only** the version string, against current master:

| benchmark | 0.8.0 | with 0.7.0's version | delta |
|---|---|---|---|
| `parse_message` | 1.969 ms | 1.899 ms | **−3.6%** |
| `full_read` | 1.975 ms | 1.911 ms | −3.2% |
| `parse_tree` | 1.951 ms | 1.876 ms | −3.9% |
| `parse_many` | 15.579 ms | 15.083 ms | −3.2% |
| `parse_lazy_all_attachments` | 2.067 ms | 1.941 ms | −6.1% |
| `parse_metadata` | 0.516 ms | 0.516 ms | −0.0% |
| `mail_parser` (control) | 13.730 ms | 13.737 ms | +0.0% |

So #197's +3.8–4.4% was real and symmetric with this: **0.8.0 as published is ~3.5% slower on the decode path than the same code carrying the previous version string.** Nothing in the crate reads `CARGO_PKG_VERSION` — verified — so this is code layout and nothing else. `parse_metadata` is untouched, which fits: it never decodes.

The useful conclusion is not "revert the version". It is that **every release cut rolls this dice**, so a version-bump PR's gate reading will land a few percent either way above the noise floor and that is expected rather than a regression. Recorded as an issue so the next person reading such a run is not misled, and so the stabilisation options have somewhere to live.

Closing; the branch was never for merging.

Issue #204, opened 2026-08-27
Code layout can swing the hot path 0-96% depending on the runner's CPU


kurok:
Measured, not suspected. Reverting **only** `Cargo.toml`'s version string from `0.8.0` to `0.7.0` against otherwise-identical master:

| benchmark | 0.8.0 | 0.7.0's version string | delta |
|---|---|---|---|
| `parse_message` | 1.969 ms | 1.899 ms | **−3.6%** |
| `full_read` | 1.975 ms | 1.911 ms | −3.2% |
| `parse_tree` | 1.951 ms | 1.876 ms | −3.9% |
| `parse_many` | 15.579 ms | 15.083 ms | −3.2% |
| `parse_lazy_all_attachments` | 2.067 ms | 1.941 ms | −6.1% |
| `parse_metadata` | 0.516 ms | 0.516 ms | −0.0% |
| controls | | | ≤ 0.1% |

Interleaved, median of three rounds, same runner ([the probe run](https://github.com/namecheap/fast_mail_parser/pull/203)). It is symmetric with what the release cut itself measured (#197: **+3.8% to +4.4%**), which is the same effect seen from the other side.

## Why this happens

**Nothing in the crate reads `CARGO_PKG_VERSION`** — verified by grep across `src/` and the Python package. The version reaches the compiler as crate metadata, which feeds symbol naming and therefore code layout. With `lto = "fat"` and `codegen-units = 1`, layout is decided across the whole graph at once, so a metadata change can move the hot loop.

`parse_metadata` being unmoved fits exactly: it walks parts and never calls `get_body_raw`, so it does not sit in the affected code.

## Why it matters, and what it is not

It is **not** a regression to revert — you cannot ship 0.7.0's version string forever. The consequence is about reading the gate:

**Every release cut rolls this dice.** A version-bump PR will show a few percent either way, above the noise floor, and that is expected. I merged #197 reading only "no failures" and so shipped ~3.5% without noticing; the next person should know the signal is there and what it means.

It also bounds what the gate's 7% threshold can honestly police: a change that genuinely costs 3% is indistinguishable from a change that merely moved the version.

## What could actually be done

- **Nothing.** ~3.5% on one of two possible outcomes per release, for free. Defensible.
- **PGO.** A profile-guided build would pin the hot layout instead of leaving it to symbol-name luck. Real work: a profile-collection step in the wheel matrix, and the profile has to be regenerated as the code changes.
[...]

PR #206, opened 2026-08-27
PROBE (do not merge): is the version effect a lottery?

kurok:
Diagnostic only — **draft, not for merge**. Follows #204.

#204 established that reverting 0.8.0's version string to 0.7.0 makes the decode path **3.2–3.9% faster**, with `parse_metadata` flat and controls under 0.1%. What it could not say is whether the effect is *arbitrary per version* or *monotonic in it*.

This bumps to **0.8.1** against master's 0.8.0, changing nothing else:

- **near −3.5%** → per-version lottery; 0.8.0 drew badly, and a patch release would recover it for free
- **near 0%** → 0.8.0 and 0.8.1 land alike, so recovery needs an actual layout lever rather than a version
- **near +3.5%** → something stranger, and worth more digging than a probe

Either way it goes on #204 and the branch is closed. Reading `parse_metadata` alongside is the control here: it never decodes, so it should stay flat whatever the decoding benchmarks do.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,7 +3,7 @@ name = "fast_mail_parser"
 description = "Very fast Python library for .eml files parsing."
 edition = "2021"
 rust-version = "1.83"
-version = "0.8.0"
+version = "0.8.1"
 authors = ["Andrii Sokyrko <wartwvister@gmail.com>"]
 readme = "Readme.md"
 # Both fields: `license` is the SPDX expression tooling reads (cargo-deny warns


kurok:
Answered, and the answer is more interesting than the question. Closing.

Bumping 0.8.0 → 0.8.1 measured **+27% on the decode path and +96% on metadata mode** on the first run, then **±0.4% on a re-run of the identical commit**. The `.so` hashes are the same in both attempts (`b93664d4…` head, `66dc7783…` base), so the *same two binaries* produced both verdicts.

It is not noise. The per-round values on the failing run are tighter than anything else measured today:

| | head | base |
|---|---|---|
| `parse_message` | 2.593 / 2.589 / 2.585 | 2.040 / 2.041 / 2.038 |
| `parse_metadata` | 1.135 / 1.135 / 1.135 | 0.577 / 0.578 / 0.578 |

Three interleaved rounds, spread under 0.3%, and a stable +27%/+96%. The controls stayed within 3.8%. That run's absolutes were ~1.8x higher throughout (control 12.7 ms vs 7.2 ms), so it was a slower runner — a different CPU.

**So the layout difference is real, and its cost is microarchitecture-dependent: nothing on one CPU, +96% on another.** Which is a far more useful statement than the one I put on #204, and I am correcting that issue accordingly.

PR #210, opened 2026-08-27
Extend mode= to parse_many and parse_email_tree (#202)

### LINKED ISSUE #202: Extend mode= to parse_many and parse_email_tree
Follow-up from #97, which delivered `mode="metadata"` and `mode="lazy"` on `parse_email` only. Its design notes assume the modes compose with the other two entry points, and they do not yet.

## What is missing

```python
parse_many(payloads, mode="metadata")      # TypeError today
parse_email_tree(payload, mode="lazy")     # TypeError today
```

Both are the cases where the modes are worth the most, which is why this is worth doing rather than noting:

- **`parse_many(..., mode="metadata")`** is the mailbox sweep the mode was built for. Today a caller who wants headers for ten thousand messages has to choose between the batch API and the mode, and the batch API is where the per-call overhead saving lives (#96: ~12x at 2000 × 0.8 KiB).
- **`parse_email_tree(..., mode="lazy")`** is the forensics case: walk the structure of a large message, decode the one part you want. Today the tree decodes every leaf eagerly.

## Notes for whoever picks it up

- The overload pattern from #185 is the shape to follow: the mode picks the return type statically, so no existing caller's types change. That needs `PyLazyMimePart` / a metadata tree node, which is the bulk of the work.
- `parse_many` returning a heterogeneous list per mode wants thought: the mode is uniform across the batch, so overloading on the literal works, but `raise_on_error` and `strict` already occupy that signature.
- **Measure before assuming it is free.** Every mode addition so far has cost the hot path 15–96% on the first attempt, from code that never executes — four times in one day. The pattern that fixed each: answer the mode in the first lines of the entry point and leave the existing body byte-for-byte unchanged, with new work behind `#[inline(never)]`. See the Performance section of `CONTRIBUTING.md`.
- `mode="lazy"` on the tree needs the same `raw_bytes` re-parse the flat lazy path uses, which is validated by a randomized sweep and by the `parse_agreement` fuzz target.

Not urgent: the modes are useful as shipped, and this is reach rather than repair.

kurok:
Closes #202.

`mode=` shipped on one entry point of three. These are the two the issue names as
where it is worth the most, and until now a caller had to choose between a mode
and the API.

```python
parse_many(payloads, mode="metadata")        # list[PyMailMetadata | ParseError]
parse_many(payloads, mode="lazy")            # list[PyLazyMail | ParseError]

parse_email_tree(payload, mode="metadata")   # PyMimePartMetadata
parse_email_tree(payload, mode="lazy")       # PyLazyMimePart
```

Both defaults are untouched, and the mode picks the return type through the
stub's overloads, so no existing caller's types change.

## Measured

767 KiB attachment-heavy fixture, median of three interleaved rounds on the CI
runner (this PR's gate run, 4 vCPU):

| | full | metadata | lazy, nothing read |
| --- | --- | --- | --- |
| `parse_email_tree` | 2.02 ms | 0.58 ms (**3.5x**) | 0.61 ms (**3.3x**) |
| `parse_many`, 8 msgs, `threads=1` | 16.16 ms | 4.68 ms (**3.5x**) | — |
| `parse_many`, 2000 × 0.8 KB | 4.67 ms | 4.11 ms (1.14x) | — |

The small-message row is the honest one to publish next to the others: small mail
is mostly headers, so there is little decoding to skip and the mode neither helps
nor hurts. The batch API's own ~12x at that size (#96) is unaffected.

Four benchmarks added for these, each guarded with `pytest.skip` so the base
build — which has no `mode=` on either function — stays green while the gate
measures this revision's benchmarks against it.

## Hot path

**The gate's verdict on this PR: no significant difference.** Worst treatment
delta **+0.8%** against an **8.2%** control noise floor; `parse_message` +0.5%,
`parse_many` -0.2%, `parse_tree` +0.7%, `parse_metadata` -0.0%. Positive means the
*base* was the slower side, so nothing here reads as a cost. Absolute floor 5.00x
vs mail-parser — pass. There is no version change in this PR, so the #204
version-string effect is not in play.

Measured locally first, which is how the shape below was arrived at rather than
guessed: interleaved A/B against `origin/master`, five rounds a side, worst
treatment delta +2.8% against a 5.0% control floor. A **null run** — the same
wheel on both sides — reported up to -2.7% on that machine, so it could not
resolve anything smaller; CI's ~0.3% within-job noise can, and did.

The pattern the four earlier regressions taught: `mode` is answered in the first
lines of both entry points and their existing bodies are byte-for-byte what they
were. Every new binding function carries `#[inline(never)]`, the two error
constructors are the existing `#[cold]` ones (`unknown_mode`,
`strict_needs_decoded_bodies`) rather than new ones, and each new mode pair shares
a single `catch_panics` call site instead of one per mode.

## Design, and why

**Two new node types, not a wider `PyMimePart`.** `PyMimePart.content` is
`bytes | None` and its `None` *means* "this node is a `multipart/*` container".
A mode where `None` also meant "not decoded yet" would make a leaf
indistinguishable from a container — the silent-wrong-answer shape #150 and
metadata mode's absent `text_plain` were both about. So `PyMimePartMetadata` has
no `content` attribute **at all**, and reports `encoded_size`: `None` in the same
place and with the same meaning as `content`'s `None`, `int` everywhere else.
Making `PyMimePart.content` lazy would separately have re-timed a shipped
attribute and moved where it raises, which #104 batches into an API-v2 window;
adding a type needs no window.

**Two rather than one**, because they differ in more than presentation. Lazy mode
retains a copy of every leaf so it can decode one later; metadata mode retains
none. On a tree that difference *is* the memory, which is the point of a sweep —
one type with an optional strategy would have had to publish `content` and
`is_decoded` on nodes that can never decode, which is the same
"attribute-that-lies" problem in a new place.

**One traversal, not three.** In the core, `NodeBody` is a three-arm enum
(`Container` / `Undecoded { encoded_size, raw: Option }` / `Decoded`) on one
`TreeNode`, and one `build_node(part, depth, defer)` builds both deferred modes,
differing in a single `bool`. Everything the modes must agree about — the
`multipart/*` and `message/rfc822` discrimination, `collect_headers`,
`part_filename`, `disposition_token`, `normalize_content_id`, `encoded_size`, both
caps, the #150 separator repair — is one piece of code. Full mode's
`MimePart::build` is deliberately left byte-for-byte untouched, so the one thing
duplicated between it and `build_node` is the two-line classification, which is
now covered by an equality assertion rather than by shared code (the same trade
`parse_email_metadata` and `parse_email_lazy` already make with `Mail::new`'s
envelope extraction).

**One batch scheduler, not three.** `parse_many_as` takes the per-message parse as
a `fn` pointer; `parse_many` is a call to it. The dynamic cursor and the order
restoration are where a bug would be subtle, so there is one of them. A `fn`
pointer rather than `impl Fn` keeps the instantiation count at the number of
result types.

**`message/rfc822` is still decoded, in every mode.** Its body *is* the embedded
message, so parsing it is what gives the node children — deferring it would defer
the *structure*, which is the one thing a tree API has to deliver eagerly, and a
mode that dropped those children would not be returning the same tree. The bytes
are already in hand, so they are published rather than dropped and decoded again:
such a node reports `is_decoded == True` from the start. Two consequences,
documented and pinned as tests: a deferred tree **can** raise `DecodeError` for a
broken embedded message, where `parse_email(mode="metadata")` never can; and a
lazy tree over a bounce holds that message decoded.

**Combinations.** `strict=True` + `mode="metadata"` raises `ValueError` on
`parse_many` with the same message it raises on `parse_email`, and before any
parsing happens; the stub's overloads reject it statically too (`mypy --strict`
verified). `strict=True` + `mode="lazy"` is honoured per slot and means what it
means everywhere else, because lazy mode finds every repair full mode finds.
`threads`, `raise_on_error`, input order and `ParseError`-in-a-slot behave
identically in all three modes. The tree has no `strict` in any mode — it has no
warnings channel to be strict about, which is the same open decision #99 left.

## Verification

- **The tree is the same tree in every mode.** Every per-node field except the
  body asserted equal to full mode's over the whole fixture + RFC corpus, and each
  leaf's lazy bytes equal to full mode's.
- **`parse_agreement` extended** with the #202 derivations: shape equality against
  full mode for both deferred trees, `encoded_size` agreement between them,
  presence agreeing with full mode's `content`, the ≤2x decode bound per node, and
  every deferred leaf decoding to exactly full mode's bytes.
- **The `raw_bytes` claim now covers a root.** Flat lazy mode defers only
  subparts; a single-part message's tree root *is* the leaf, so the retained slice
  is the whole payload. Asserted per fixture and on arbitrary input rather than
  assumed.
- `walk` accepts a node from any mode and yields nodes of the same type. The
  implementation was already duck-typed; the stub now says so with a
  value-restricted `TypeVar`, which is non-breaking — walking a `PyMimePart` still
  yields `PyMimePart`.
- 801 tests pass (up from 623); `test_contract.py`, `test_stub_matches_runtime.py`
  and `test_readme_snippets.py` updated/green; `cargo fmt --check`, `cargo clippy
  --all-targets -D warnings`, `mypy --strict`, `ruff check` all clean.

## Not verified locally

- **`cargo fuzz` was not run locally**: no nightly and no `cargo-fuzz` on this
  machine. The target compiles, and its new invariants were exercised by a
  standalone driver over the corpus plus 38,024 deterministic mutations (bit
  flips, truncations, spliced `message/rfc822` and quoted-printable headers) — all
  held. CI's `fuzz (short pass)` is green on this branch; the deep run is where a
  real exploration happens.
- **Free-threaded builds.** `PyLazyMimePart`'s cache is the same `OnceLock`-on-the-
  Python-object shape `PyLazyAttachment` uses, so the #101 audit's invariant is
  untouched: nothing was added to the parsing core that is shared or mutable. The
  concurrency tests for the new type are the single-threaded ones only; the
  existing concurrent tests cover the identical mechanism on `PyLazyAttachment`.
- **`encoded_size` was not added to `PyMimePart`.** It would be additive and is
  arguably useful there too, but it is not what #202 asks for.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,78 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **`mode=` on `parse_many` and `parse_email_tree`** (#202), honouring the same
+  `"full"` / `"lazy"` / `"metadata"` values `parse_email` honours. The mode axis
+  shipped on one entry point of three; these are the two where it is worth the
+  most, and until now a caller had to choose between a mode and the API.
+  - **`parse_many(payloads, mode="metadata")`** is the mailbox sweep the mode was
+    built for. The batch API removes per-message overhead (#96: ~12x at
+    2000 × 0.8 KB) and metadata mode removes the decoding (#97: ~3x on an
+    attachment-heavy message); they now compose. Measured on the 767 KiB fixture,
+    8 messages, `threads=1`: full 8.96 ms, metadata 2.94 ms — **3.0x**, the same
+    ratio the single-message mode gets. On 2000 × 0.8 KB it is 2.80 ms against
+    3.04 ms, because small messages are mostly headers and there is little
+    decoding to skip.
+  - **`parse_email_tree(payload, mode="lazy")`** is the forensics case: walk the
+    structure of a large message and decode the one part you want. A full-mode
+    tree decodes every leaf, which is the wrong bill for exactly what the tree is
+    best at. `mode="metadata"` decodes nothing *and retains nothing*. On the same
+    fixture: full tree 1.10 ms, lazy with nothing read 0.38 ms, metadata
+    0.37 ms — **~2.9x**.
+  - **Two new node types, `PyLazyMimePart` and `PyMimePartMetadata`**, rather than
+    a wider `PyMimePart`. `PyMimePart.content` is `bytes | None` and its `None`
+    *means* "this node is a `multipart/*` container"; a mode where it also meant
+    "not decoded yet" would make a leaf indistinguishable from a container, which
+    is the silent-wrong-answer shape #150 and metadata mode's absent `text_plain`
+    were both about. A metadata node therefore has no `content` attribute at all
+    and reports `encoded_size` instead — `None` in the same place, and with the
+    same meaning, as `content`'s `None`. Making `PyMimePart.content` lazy would
+    also have re-timed a shipped attribute and moved where it raises, which is
+    what #104 batches into an API-v2 window; adding a type is not.
+  - **Two rather than one**, because the two differ in more than presentation:
+    lazy mode retains a copy of every leaf so it can decode one later, metadata
+    mode retains none. On a tree that difference is the memory, which is the point
+    of the sweep.
+  - **The tree is the same tree in every mode.** Every per-node field except the
+    body is asserted equal to full mode's across the whole fixture and RFC corpus,
+    and on arbitrary input by the `parse_agreement` fuzz target, which gained the
+    #202 derivations. A mode that quietly reshaped a message would be the exact
+    failure that target exists to catch, and it has caught two already.
+  - **`message/rfc822` is still decoded, in every mode.** Its body *is* the
+    embedded message, so parsing it is what gives the node children — deferring it
+    would defer the structure, which is the one thing a tree API has to deliver
+    eagerly. The bytes are published rather than dropped and decoded again, so
+    such a node has `is_decoded == True` from the start; and unlike
+    `parse_email(mode="metadata")`, a deferred tree can raise `DecodeError` for
+    such a part. Documented, and pinned as a test.
+  - **The lazy tree extends the `raw_bytes` claim to a root.** Flat lazy mode
+    defers only subparts; a single-part message's tree root *is* the leaf, so the
+    retained slice is the whole payload. That the re-parse still reproduces full
+    mode's bytes is asserted per fixture and on arbitrary input rather than
+    assumed.
+  - **Combinations that cannot mean anything are refused.** `strict=True` with
+    `mode="metadata"` raises `ValueError` on `parse_many` with the same message it
+    raises on `parse_email`, and before any parsing happens. `strict=True` with
+    `mode="lazy"` is honoured per slot and means what it means everywhere else.
+    The tree has no `strict` in any mode, because it has no warnings channel to be
+    strict about.
+  - `walk` accepts a node from any mode and yields nodes of the same type. Its
+    implementation was already duck-typed; the stub now says so with a
+    value-restricted `TypeVar`, which is non-breaking — walking a `PyMimePart`
+    still yields `PyMimePart`.
+  - **Measured, not assumed.** Interleaved A/B against `master`, three rounds a
+    side: worst treatment delta +0.5% against a 4.4% control noise floor, with
+    `parse_message` at -1.2% and `parse_many` at -0.2%. The default paths are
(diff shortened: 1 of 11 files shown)


kurok:
Verified independently before merging: 15/15 green, gate worst delta **+0.8%** against an 8.2% control floor, rebase onto current master is clean, and the checks ran on the merge ref so they exercised the combined state.

Worth noting the gate's output now carries `Measured on \`AMD EPYC 9V74 80-Core Processor, 4 vCPU\`` — that line landed in #207 this afternoon after a probe changing nothing but a version string measured +96% on one runner and ±0.4% on another with identical binaries. This PR's run is the first to have it, and it is what makes two runs comparable at all.

PR #222, opened 2026-08-28
Stop the scheduled tasks: no weekly deep fuzz, dependabot paused

kurok:
Stops everything that runs on a schedule in this repository, without removing the ability to run any of it by hand.

- **`deep-fuzz.yml`**: the Monday cron is gone; `workflow_dispatch` stays, with the same inputs and the same cold-cache seeding from the `fuzz-corpus-*` release asset. The header comment now describes an on-demand run.
- **`dependabot.yml`**: all three ecosystems set to `open-pull-requests-limit: 0` — GitHub's documented way to pause version updates while keeping the configuration, so re-enabling is deleting three lines. Security updates are unaffected (they are a separate setting).
- **CONTRIBUTING** no longer promises a weekly run; **CHANGELOG** records the change.

Not touched: CodeQL default setup (runs on push/PR; disabling code scanning is a security decision, not a scheduled-task one) and the badge on the README, which will show the last on-demand run's status. Both workflows still parse as YAML locally.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,16 +1,21 @@
 version: 2
+# Version updates are PAUSED (limit 0 on every ecosystem) since 2026-08-28: no
+# scheduled tasks run in this repository. Delete the three limit lines to resume.
+# Security updates are a separate repository setting and are unaffected.
 updates:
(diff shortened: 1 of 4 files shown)


kurok:
Opened by mistake — the request was about local session tasks, not repository schedules. No change intended here.

Issue #226, opened 2026-08-30
Borrow `str` payloads through `PyBackedStr` instead of copying them (the gate benchmark pays this memcpy on every iteration)


kurok:
## Problem

Every `parse_email(str)`, `parse_email_tree(str)` and every `str` slot of `parse_many` first duplicates the whole message into a fresh `Vec<u8>` so the parse can run with the GIL released. `bytes` payloads stopped being copied in #169, but that PR left `str` copied on the stated grounds that the limited API offers no UTF-8 buffer to borrow. That is not true for the ABI this crate builds: with `pyo3/abi3-py311` the `Py_3_10` cfg is on, `PyString::to_str()` is the zero-copy `PyUnicode_AsUTF8AndSize` path (the code already calls it one line before the copy), and PyO3's `PyBackedStr` wraps exactly that borrow in a `Send + Sync` handle that keeps the `str` alive, the same way `PyBackedBytes` does for `bytes`.

This matters because the CI performance gate measures the `str` path. `tests/benchmark/test_read_message.py::test__fast_mail_parser___parse_message` is fed the `large_message` fixture as a `str` (785,774 bytes, pure ASCII), so every timed iteration includes a ~785 KiB `malloc` + `memcpy` + `free` that does no parsing. For `mode="metadata"` (0.030 ms with `bytes` on the M4) the copy is of the same order as the entire parse.

## Evidence

All citations are master `73f56da`.

- `src/fast_mail_parser.rs:1209-1211`:
  ```rust
  if let Ok(text) = obj.cast::<PyString>() {
      if let Ok(text) = text.to_str() {
          return Ok(Payload::Owned(text.as_bytes().to_vec()));
  ```
  `to_str()` already borrows the UTF-8 buffer; `to_vec()` then copies all of it.
- `src/fast_mail_parser.rs:1183-1187` (doc comment on `Payload`): "`str` is a different case and still gets copied. Under the limited API, obtaining UTF-8 from a `str` means asking CPython to encode it, which allocates -- there is no buffer to borrow." This is the premise the code rests on, and it is false for this build.
- `Cargo.toml` `[features]`: `default = ["pyo3/extension-module", "pyo3/abi3-py311"]` -- the 3.11 baseline sets `Py_3_10`.
[...]

PR #243, opened 2026-09-17
Share attachment objects and their bytes instead of copying per read (#227)

### LINKED ISSUE #227: Full-mode attachments are deep-copied on every `mail.attachments` read and again on every `.content` read; give PyAttachment/PyMimePart the lazy types' `Py<_>` + `OnceLock<Py<PyBytes>>` shape
## Problem

In the default (full) mode, reading an attachment's bytes from Python costs two full copies of the decoded payload, and the two public attachment-bearing types (`PyAttachment`, `PyMimePart`) behave differently from their lazy siblings (`PyLazyAttachment`, `PyLazyMimePart`) on the same attribute.

The first copy is hidden in `PyMail.attachments`. The field is `#[pyo3(get)] pub attachments: Vec<PyAttachment>`, and `PyAttachment` carries `#[derive(Clone)]` with its decoded bytes inline as `content: Vec<u8>`. PyO3 0.29 has no by-reference conversion for a `#[pyclass]` value, so this getter falls back to the clone path: every read of `mail.attachments` clones the whole `Vec<PyAttachment>` -- a `memcpy` of every attachment's decoded bytes -- and allocates fresh `PyAttachment` objects. `mail.attachments[0] is mail.attachments[0]` is `False`. The second copy is the `content` getter itself: `PyBytes::new(py, self.content.as_slice())` builds a new `bytes` on each read, so `a.content is a.content` is also `False`.

That makes the library's own documented idiom (`by_cid = {a.content_id: a for a in mail.attachments ...}` then `len(attachment.content)`, README.md:266-275) and the published comparison benchmark (`_fast_mail_parser_full`, which reads `mail.attachments` and each `a.content` once) pay two copies per attachment on top of the decode. The fixture `tests/data/large_message.eml` is 767 KiB and "99% attachment by decoded content" (two base64 attachments), so on `test__fast_mail_parser___full_read` the copies are a measurable fraction of the 0.225 ms (M4) / 0.604 ms (EPYC) parse. Lazy mode already solved both halves in 0.9.0 (#97): `PyLazyMail.attachments` is `Vec<Py<PyLazyAttachment>>` handed out with `clone_ref`, and `PyLazyAttachment.content` is a `OnceLock<Py<PyBytes>>` that publishes one `bytes` object. The full-mode types simply predate that pattern.

## Evidence

- `src/fast_mail_parser.rs:741-746` -- `#[pyclass(skip_from_py_object)] #[derive(Clone)] pub struct PyAttachment { #[pyo3(get)] pub mimetype: String, pub content: Vec<u8>, ...`. The decoded bytes live inline in a `Clone` pyclass.
- `src/fast_mail_parser.rs:822-823` -- `#[pyo3(get)] pub attachments: Vec<PyAttachment>,` on `PyMail` (struct at 789). A `#[pyo3(get)]` on a `Vec<pyclass>` field.
- PyO3 0.29.2, `src/impl_/pyclass.rs:1333-1350` ("Base case attempts to use IntoPyObject + Clone") and `1414-1438` -- `pyo3_get_value_into_pyobject`: `let value = ...as_ref(); value.clone().into_py_any(py)`. The by-reference path (`1304-1323`, requires `for<'a> &'a FieldT: IntoPyObject`) is not available because the `#[pyclass]` macro only emits `IntoPyObject for PyAttachment` by value (`pyo3-macros-backend-0.29.2/src/pyclass.rs:2778-2790`). So every `mail.attachments` read deep-clones every attachment. The `#[derive(Clone)]` at 742 exists only to satisfy this path: `grep` finds no other clone of `PyAttachment`.
- `src/fast_mail_parser.rs:765-769` -- `#[getter] fn content<'py>(&self, py: Python<'py>) -> Bound<'py, PyBytes> { PyBytes::new(py, self.content.as_slice()) }`. A second full copy per `.content` read.
- `src/fast_mail_parser.rs:352-353, 372-377` -- `PyMimePart { ... pub content: Option<Vec<u8>>, pub children: Vec<Py<PyMimePart>> }` and `fn content(...) -> Option<Bound<PyBytes>> { self.content.as_ref().map(|bytes| PyBytes::new(py, bytes.as_slice())) }`. Same per-read copy; note `children` already uses `Py<_>` and a `clone_ref` getter (381-383), so the tree type is half-converted.
- `src/fast_mail_parser.rs:1091, 1098-1101` -- `pub attachments: Vec<Py<PyLazyAttachment>>` with the comment "Held as Python objects rather than Rust values, which is what makes the cache mean anything", and `fn attachments(...) { PyList::new(py, self.attachments.iter().map(|part| part.clone_ref(py))) }`. The target shape for `PyMail.attachments`.
- `src/fast_mail_parser.rs:963, 983-988, 1047-1055` -- `content: OnceLock<Py<PyBytes>>`; `if let Some(cached) = self.content.get() { return Ok(cached.bind(py).clone()); }`; `let bytes = PyBytes::new(py, decoded.as_slice()).unbind(); Ok(self.content.get_or_init(|| bytes).bind(py).clone())`. The target shape for `content`. (`PyLazyMimePart` repeats it at 551-560, 590-596.)
- `src/fast_mail_parser.rs:954-962` -- the reason lazy mode decodes *outside* `get_or_init`: holding the cell's lock across a `detach` can deadlock. Full mode never detaches in the getter (the bytes are already decoded), so it may initialise inside `get_or_init` safely.
- `src/fast_mail_parser.rs:1137-1142` -- `PyLazyMail::from_lazy(py, mail)` builds `Py::new(py, PyLazyAttachment::from_lazy(part))` per attachment with the GIL held; `PyMail::from_mail(mail)` at 882-910 has no `py` and stores plain values. Both callers of `from_mail` already hold `py`: `parse_email_inner` (1403-1418, after `py.detach`) and `parse_many_inner` (1555-1567).
- `tests/benchmark/test_read_message.py:87-95, 137-141` -- `_fast_mail_parser_full` returns `[(a.mimetype, a.filename, a.content) for a in mail.attachments]`; it is `test__fast_mail_parser___full_read`, the number published in README.md:129. Two copies per attachment inside the timed body.
- `tests/test_lazy_mode.py:16-18, 174-193, 253-290` and `tests/test_tree_modes.py:290-293` -- the identity contract (`a.content is a.content`, `mail.attachments[0] is mail.attachments[0]`, one object across racing threads) is tested for the lazy types only.
- `tests/test_contract.py:66-72, 154-160` -- the frozen `PyAttachment` surface is the attribute *names* and `isinstance(attachment.content, bytes)`; identity and cost are not frozen. README.md:499-501 ("`PyAttachment.content` is exactly what it was") is about *where the decode happens and where `DecodeError` raises*, both untouched here.
- Cargo.toml `[features]`: `default = ["pyo3/extension-module", "pyo3/abi3-py311"]` -- `py-clone` is off, so `Py<T>` is not `Clone`; the struct cannot keep `#[derive(Clone)]` once it holds a `Py<PyBytes>`, which is what forces `PyMail.attachments` onto the `Vec<Py<_>>` shape.
- Known profile (maintainer's M4 sample of a full parse): base64 decode 53%, whitespace strip 27%, memmem 7.5%. The copies here are *outside* `parse_email` and so absent from that profile; they land on whichever caller reads attachments -- which is every consumer of the default mode that wants the bytes.

## Proposal

Make the two full-mode types hand out the same objects the lazy ones do. No new dependency. No change to attribute names, types, decode timing, or where errors raise.

1. `PyAttachment`: drop `#[derive(Clone)]`; keep `content: Vec<u8>` (decoded in `parse_email`, as today); add `content_py: OnceLock<Py<PyBytes>>`; the getter builds the `bytes` on first read and returns the cached object afterwards.
2. `PyMail.attachments`: `Vec<Py<PyAttachment>>` with an explicit `#[getter]` using `clone_ref`, exactly as `PyLazyMail` at 1098-1101. `from_mail` takes `py` and returns `PyResult<Self>`.
3. `PyMimePart`: keep `content: Option<Vec<u8>>`, add `content_py: OnceLock<Py<PyBytes>>`, same getter shape returning `Option`.

```rust
#[pyclass(skip_from_py_object)]
pub struct PyAttachment {
    #[pyo3(get)]
    pub mimetype: String,
    /// Decoded by `parse_email`, so a value rather than a deferred decode; the
    /// lazy type is where that trade is made.
    pub content: Vec<u8>,
    /// The `bytes` handed to Python, built on first read and then shared, so
    /// `a.content is a.content` as it is for `PyLazyAttachment`. Unlike the
    /// lazy cell this may initialise inside `get_or_init`: `PyBytes::new` never
    /// releases the GIL and cannot fail, so the deadlock the lazy note
    /// describes cannot arise here.
    content_py: OnceLock<Py<PyBytes>>,
    // filename, content_id, disposition unchanged
}

#[pymethods]
impl PyAttachment {
    #[getter]
    fn content<'py>(&self, py: Python<'py>) -> Bound<'py, PyBytes> {
        self.content_py
            .get_or_init(|| PyBytes::new(py, &self.content).unbind())
            .bind(py)
            .clone()
    }
}

// PyMail
pub attachments: Vec<Py<PyAttachment>>,

#[getter]
fn attachments<'py>(&self, py: Python<'py>) -> PyResult<Bound<'py, PyList>> {
    PyList::new(py, self.attachments.iter().map(|a| a.clone_ref(py)))
}

impl PyMail {
    pub(crate) fn from_mail(py: Python<'_>, mail: mail_parser::Mail) -> PyResult<Self> {
        let attachments = mail.attachments.into_iter()
            .map(|a| Py::new(py, PyAttachment::from_attachment(a)))
            .collect::<PyResult<Vec<_>>>()?;
        Ok(Self { attachments, /* rest unchanged */ })
    }
}
```

Do **not** convert `Vec<u8>` to `PyBytes` eagerly in `from_mail`: that would put a copy of every attachment into `parse_email`'s own timed path. The `Py::new` per attachment does land in that path (GIL held, after `detach`) -- it is the same cost lazy mode already pays at 1137-1142 and is expected to be sub-microsecond on the two-attachment fixture, but it is inside the gate and must be measured.

Behaviour that must stay identical: attribute set of `PyAttachment`/`PyMail`/`PyMimePart` (test_contract.py); `content` is `bytes` (or `None` for a container); decoding still happens in `parse_email`/`parse_email_tree` and `DecodeError` still raises there, not from the getter; `warnings`, `strict=`, `parse_many` slot semantics unchanged. The free-threading audit boundary at `src/mail_parser.rs:8-26` is untouched: the cell lives in the binding layer on the Python object, exactly like the lazy one (the audit note at 27-33 already describes that as the accepted shape).

Memory: today a caller holding `a = mail.attachments[0]` and `a.content` has three copies of the payload alive (PyMail's `Vec`, the clone inside `a`, the `bytes`). After: two (the `Vec` and the `bytes`), and one until `content` is first read. Dropping the `Vec` after the `bytes` exists is possible with a `&mut self` getter and an enum, but it is a different design from the lazy types; leave it out of scope and note it in the PR.

1. `src/fast_mail_parser.rs`:
   - `PyAttachment` (741-770): remove `#[derive(Clone)]`, add `content_py: OnceLock<Py<PyBytes>>` (initialised `OnceLock::new()` in `from_attachment`, 773-781), rewrite the `content` getter as sketched. Update the field doc; cross-reference the lazy note at 954-962 and say why in-`get_or_init` init is fine here.
   - `PyMail` (789-838): `attachments: Vec<Py<PyAttachment>>`, remove `#[pyo3(get)]` from it, add a `#[getter] fn attachments` mirroring 1098-1101 (keep the RFC 2183 doc comment on the getter). Change `from_mail` (882-910) to `from_mail(py, mail) -> PyResult<Self>` and update both callers: `parse_email_inner` 1417 (`Ok(PyMail::from_mail(py, mail)?)`) and `parse_many_inner` 1557 (`let mail = PyMail::from_mail(py, mail)?;`). `strict_rejection(&mail.warnings)` at 1558-1561 is unaffected.
   - `PyMimePart` (331-353, 372-377): add `content_py: OnceLock<Py<PyBytes>>`, getter returns `Option<Bound<PyBytes>>` via `self.content.as_ref().map(|v| self.content_py.get_or_init(|| PyBytes::new(py, v).unbind()).bind(py).clone())`. Initialise in `from_part` (396-415). `__repr__` unchanged.
   - `PyAttachmentMetadata` (197), `PyAddress` (720), `ParseWarning` (672) keep their `#[derive(Clone)]`: they are small and their getters legitimately go through the clone path. Do not widen the change to them.
2. Compile check on the derive: with `py-clone` off, `Py<PyBytes>` is not `Clone`, so a leftover `#[derive(Clone)]` on `PyAttachment` fails to build -- that is the signal you converted the right field. Do **not** enable `py-clone` to get around it.
3. Tests (the oracle):
   - `tests/test_attachments.py`: add `test__repeated_content_reads_return_the_same_object` (`a.content is a.content`, twice) and `test__the_attachment_list_hands_back_the_same_objects` (`mail.attachments[0] is mail.attachments[0]`), mirroring tests/test_lazy_mode.py:174-193. Add a barrier-synchronised thread test mirroring tests/test_lazy_mode.py:253-290: N threads take `mail.attachments[0].content` for the first time simultaneously and all `id()`s agree.
   - `tests/test_mime_tree.py`: add `leaf.content is leaf.content` for a full-mode leaf (mirror tests/test_tree_modes.py:290-293) and `root.content is None` still holds for containers (91, 97 already assert this).
   - Must pass unchanged: `tests/test_contract.py` (frozen `EXPECTED_ATTACHMENT_ATTRS` at 66-72 -- the new `OnceLock` field has no `#[pyo3(get)]`, so `_public_attrs` must not grow), `tests/test_stub_matches_runtime.py`, `tests/test_lazy_mode.py` (its full-vs-lazy equality tests at 87-150 are the cross-check that `content` bytes are unchanged), `tests/test_stdlib_parity.py`, `tests/test_parse_many.py`, `tests/test_tree_modes.py`, `tests/test_decode_errors.py` (DecodeError still raised by `parse_email`, not the getter).
4. Stubs: `fast_mail_parser/__init__.pyi` needs no signature change (`self.content = content`, `self.attachments = attachments` at 314, 480 stay). Optionally add one sentence to the `PyAttachment` and `PyMimePart` docstrings (286-302, 28-48): repeated reads of `content` return the same `bytes` object. `mypy --strict` must stay green.
5. Docs: README.md:499-501 stays true; append a sentence noting that full-mode `content` now also returns one shared object, so the identity line at 459 holds in every mode. `CHANGELOG.md` `## [Unreleased]` -> `### Changed`: one entry stating the two removed copies, the identity guarantee, and the memory change (3 -> 2 copies held).
6. Traps: no `vendor/` change, so `PATCH.md` and the vendored test run are not needed; abi3 is unaffected (`OnceLock`, `Py<T>` are already used by the lazy types under the same feature set); do not run `cargo` inside `vendor/mailparse` without `--target-dir` elsewhere (the sdist include glob sweeps `target/`); `PyMail` is `#[pyclass]` without `frozen` -- that is fine, `PyLazyMail` is the same and holds `Py<_>` fields; no `__traverse__` needed (no reference cycles: attachments do not point back at the mail).

## Harness -- how to prove it

Build both sides in separate trees and venvs, the way `.github/workflows/test.yml` (395-449) does, then run **this revision's** benchmark file against both builds, interleaved, three rounds each:

```bash
# base = master 73f56da (or the merge-base), head = the branch
git worktree add /path/to/base 73f56da
python -m venv /path/to/venv-base && CARGO_TARGET_DIR=/path/to/target-base \
  /path/to/venv-base/bin/pip install -q "/path/to/base[test]"
python -m venv /path/to/venv-head && CARGO_TARGET_DIR=/path/to/target-head \
  /path/to/venv-head/bin/pip install -q ".[test]"
# (or `uv pip install maturin` + `maturin develop --release` in each venv)

for round in 1 2 3; do
  for side in head base; do
    /path/to/venv-$side/bin/pytest -q tests/benchmark \
      --benchmark-min-rounds=25 --benchmark-json=$side-$round.json
  done
done

  --a-label "this revision" --b-label base \
  --informational test__threaded___ \
python .github/scripts/check_benchmark.py head-1.json
```

Thresholds and expectations (`ab_median.py` reports medians of per-benchmark minimums; the control benchmarks `test__mail_parser___*`, `test__mailparser_lib___*`, `test__stdlib_email___*` are the noise floor and must stay within a few percent):

- No benchmark may regress more than 7%. In particular the gate pair `test__fast_mail_parser___parse_message` (and `_strict`, `parse_many`, `parse_tree`) must stay within noise: they gain one `Py::new` per attachment (two on the fixture), which is expected to be well under 1% of 0.225 ms but is the thing to watch.
- `test__fast_mail_parser___full_read` is **expected to improve**: it loses one `memcpy` of every attachment's decoded bytes plus one `Vec<PyAttachment>` clone per call. On the ~767 KiB fixture that is on the order of tens of microseconds against a ~0.25 ms body, so expect a single-digit to low-double-digit percent improvement; report the measured number, not this estimate.
- New benchmark (add to `tests/benchmark/test_read_message.py`, no skip guard needed -- the API is identical on both sides): `test__fast_mail_parser___attachment_reread` -- parse `large_message` **outside** the timed call, then time `lambda: [len(a.content) for a in mail.attachments]`. On base this is two copies per attachment per call; on head the first call materialises the `bytes` and every later call is two refcount bumps per attachment. Expected: head at least 10x faster than base on this benchmark; report the ratio in the PR body.
- Memory: the existing `tests/test_payload_memory.py` covers payload borrowing only; a resident-memory assertion is not required here, but the PR body should state the 3 -> 2 copies argument.

The M4 is not a proxy for x86: the fixture's loops have shown up to 96% layout swings on Zen (#204), so after the local A/B, confirm the gate on the PR's own CI run (interleaved, EPYC) and, if any full-mode number moves more than the control floor in either direction, run `.github/workflows/toolchain-ab.yml` via `workflow_dispatch` on the branch for a second x86 reading. Local before/after `ab_median.py` output goes in the PR body.

Full local CI surface, all green:

```bash
cargo fmt --all -- --check
ruff check .
```

(`cargo test --manifest-path vendor/mailparse/Cargo.toml --target-dir /tmp/vendor-target` is only needed if `vendor/` is touched; this change must not touch it.)

## Acceptance criteria

- [ ] `PyAttachment` has no `#[derive(Clone)]`; its `content` getter returns a `Py<PyBytes>` cached in a `OnceLock` on the object; `parse_email(...).attachments[0].content is parse_email(...).attachments[0].content` -- tested in `tests/test_attachments.py`.
- [ ] `PyMail.attachments` is `Vec<Py<PyAttachment>>` with a `clone_ref` getter; `mail.attachments[0] is mail.attachments[0]` -- tested.
- [ ] `PyMimePart.content` returns the same `bytes` object on repeated reads and still `None` for containers -- tested in `tests/test_mime_tree.py`.
- [ ] Concurrent first reads of one full-mode attachment's `content` from several threads observe one object -- tested (mirror of `tests/test_lazy_mode.py:253-290`).
- [ ] `tests/test_contract.py`, `tests/test_stub_matches_runtime.py`, `tests/test_lazy_mode.py`, `tests/test_stdlib_parity.py`, `tests/test_decode_errors.py` pass **unmodified**; `pytest tests --ignore=tests/benchmark` is green.
- [ ] `DecodeError` for a broken attachment is still raised by `parse_email`/`parse_email_tree`, never by the getter (existing tests at `tests/test_lazy_mode.py:323-326` and `tests/test_decode_errors.py` are the oracle).
- [ ] New benchmark `test__fast_mail_parser___attachment_reread` exists and its head/base ratio is reported in the PR body.
- [ ] Interleaved local A/B (3 rounds, `ab_median.py`, threshold 7%): no benchmark regresses beyond the threshold; control benchmarks within their floor; `test__fast_mail_parser___parse_message` within noise; `test__fast_mail_parser___full_read` delta reported.
- [ ] CI performance gate on the PR is green on the x86 runner.
- [ ] `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings -W clippy::cast_possible_truncation`, `mypy --strict fast_mail_parser/`, `ruff check .` all green.
- [ ] `CHANGELOG.md` `[Unreleased]` entry added; README note added next to lines 499-501; `__init__.pyi` signatures unchanged.
- [ ] `vendor/`, `Cargo.toml` dependencies and features (no `py-clone`), and `src/mail_parser.rs` are untouched.

## Out of scope

- Dropping the Rust `Vec<u8>` once the `bytes` exists (would need a `&mut self` getter and an enum state; different design from the lazy types -- consider separately if resident memory of full mode matters).
- Changing `PyAttachmentMetadata`, `PyAddress`, or `ParseWarning` getters (small values; the clone path is fine there).
- Decoding directly into a `PyBytes` buffer (the decode runs with the GIL released, so it cannot allocate a Python object).
- Anything in `vendor/mailparse` or the decode pipeline itself (base64 decode / whitespace strip are separate work).

## Related issues

- #231 — same `OnceLock<Py<_>>` pattern; its Out of scope hands this case here
- #239

kurok:
Closes #227.

Depends on nothing; branched from `master` (`73f56da`), independent of #242.

## What changed

Reading an attachment's bytes in the default mode cost **two full copies** of the decoded payload, and neither shows up in a profile of `parse_email` — both land on the caller.

The first copy hid in `PyMail.attachments`: `#[pyo3(get)]` on a `Vec<PyAttachment>`, and PyO3 0.29 has no by-reference conversion for a `#[pyclass]` value, so the getter falls back to the clone path. Every read of `mail.attachments` deep-cloned the vector — a `memcpy` of every attachment's decoded bytes — and handed back fresh objects, so `mail.attachments[0] is mail.attachments[0]` was `False`. The `#[derive(Clone)]` on `PyAttachment` existed only to feed that path. The second copy was the `content` getter building a new `bytes` per read.

So the library's own documented idiom — `by_cid = {a.content_id: a for a in mail.attachments}`, then read `a.content` — paid two copies of every payload per pass, on top of the decode it had already done.

Lazy mode fixed both halves in #97, and the tree type was half-converted already (`children` has used `Py<_>` + `clone_ref` since #99). This gives the two full-mode types the same shape:

| | before | after |
|---|---|---|
| `PyMail.attachments` | `Vec<PyAttachment>` + `#[pyo3(get)]` | `Vec<Py<PyAttachment>>` + `clone_ref` getter |
| `PyAttachment.content` | `PyBytes::new` per read | `OnceLock<Py<PyBytes>>`, published once |
| `PyMimePart.content` | `PyBytes::new` per read | same, still `None` for a container |

`PyAttachment` loses `#[derive(Clone)]` — which is the compiler's own check that the right field moved, since `py-clone` is off and `Py<PyBytes>` is not `Clone`. `PyAttachmentMetadata`, `PyAddress` and `ParseWarning` keep theirs: they're small and the clone path is right for them.

One design note: unlike the lazy cell — which decodes *outside* `get_or_init` to avoid holding the cell's lock across a `detach` — this one initialises inside it. The bytes are already decoded, and `PyBytes::new` neither releases the GIL nor fails, so the reentrancy that note warns about cannot arise. The comment says so at the field.

`from_mail` now takes `py` and returns `PyResult<Self>` for the one `Py::new` per attachment. That does land inside `parse_email`'s own timed path, which is why it's measured below. Nothing is copied there: the decoded `Vec<u8>` moves into the object and only becomes `bytes` when `content` is first read.

## Contract

Unchanged. Attribute sets, types, decode timing and where `DecodeError` raises are all the same, and `tests/test_contract.py`, `tests/test_stub_matches_runtime.py`, `tests/test_lazy_mode.py`, `tests/test_stdlib_parity.py` and `tests/test_decode_errors.py` pass **unmodified**. The new `OnceLock` field carries no `#[pyo3(get)]`, so the frozen public attribute set does not grow.

What is *added* is a guarantee full mode did not previously make, now tested the way #97 tested it for lazy mode:

- `a.content is a.content`
- `mail.attachments[0] is mail.attachments[0]`
- sixteen barrier-synchronised threads taking one attachment's `content` for the first time simultaneously all observe **one** object
- the same for a MIME-tree leaf, with `root.content is None` still holding for containers

**All four of these fail against master `73f56da` and pass here** — I ran them both ways rather than trusting that they were testing something.

Memory: a caller holding `a = mail.attachments[0]` and `a.content` kept three copies of the payload alive (the `Vec`, the clone inside `a`, the `bytes`); now two, and one until `content` is first read. Dropping the Rust `Vec<u8>` once the `bytes` exists is out of scope — it needs a `&mut self` getter and an enum, which is a different design from the lazy types.

## Numbers

Apple M4 (10 vCPU), rustc 1.98.0, CPython 3.12. Two *independent* 5-round interleaved A/Bs pooled to 10 rounds per side (`.github/scripts/ab_median.py`). Pure-Python controls within **0.9%**.

|---|---|---|---|
| `test__fast_mail_parser___attachment_reread` *(new)* | 12.833 µs | **0.083 µs** | **~156x** |
| `test__fast_mail_parser___full_read` | 0.250 ms | **0.243 ms** | −2.9% |

`attachment_reread` parses once outside the timed body, so it measures only what a consumer pays to *get at* the bytes — one `attachments` read plus one `.content` read per attachment. That is where the two copies were, and it is now essentially free. `full_read` includes the parse, which is decode-dominated (base64 53%, whitespace strip 27%), which is why the same absolute saving reads as 2.9% there.

Every other benchmark stayed inside the control noise floor.

**One thing worth flagging about the method:** `parse_tree` read **+2.4%** (slower) on the first A/B, above that run's 1.7% noise floor, with visibly bimodal rounds on both sides. It read **−0.5%** on a second, independent A/B. Pooled, it is −0.2%. So that was the measurement, not the change — but #204 is exactly why this repo re-runs before believing a single one, and I did not want a reader to find the first table and wonder.

<details><summary>Full pooled A/B report (10 rounds/side)</summary>

Median of 10 interleaved rounds per side; each value is a benchmark's minimum. Positive delta = `master` is slower.

| Benchmark | share-attachments | master | Delta | |
|---|---|---|---|---|
| `test__fast_mail_parser___attachment_reread` | 0.000 ms | 0.013 ms | +15614.6% |  |
| `test__fast_mail_parser___full_read` | 0.243 ms | 0.250 ms | +2.9% |  |
| `test__fast_mail_parser___parse_lazy_all_attachments` | 0.255 ms | 0.257 ms | +0.8% |  |
| `test__fast_mail_parser___parse_lazy_untouched` | 0.043 ms | 0.043 ms | +0.0% |  |
| `test__fast_mail_parser___parse_many` | 1.960 ms | 1.973 ms | +0.7% |  |
| `test__fast_mail_parser___parse_many_metadata` | 0.237 ms | 0.237 ms | +0.1% |  |
| `test__fast_mail_parser___parse_message` | 0.245 ms | 0.244 ms | -0.5% |  |
| `test__fast_mail_parser___parse_message_strict` | 0.239 ms | 0.241 ms | +0.6% |  |
| `test__fast_mail_parser___parse_metadata` | 0.030 ms | 0.030 ms | +0.2% |  |
| `test__fast_mail_parser___parse_tree` | 0.238 ms | 0.238 ms | -0.2% |  |
| `test__fast_mail_parser___parse_tree_lazy_untouched` | 0.041 ms | 0.042 ms | +0.3% |  |
| `test__fast_mail_parser___parse_tree_metadata` | 0.031 ms | 0.031 ms | -0.1% |  |
| `test__mail_parser___parse_message` | 5.484 ms | 5.487 ms | +0.1% | control |
| `test__mailparser_lib___full_read` | 5.754 ms | 5.804 ms | +0.9% | control |
| `test__stdlib_email___full_read` | 7.423 ms | 7.437 ms | +0.2% | control |
| `test__threaded___parse_many` | 1.003 ms | 1.009 ms | +0.7% | informational |
| `test__threaded___parse_many_metadata_small` | 2.511 ms | 2.549 ms | +1.5% | informational |
| `test__threaded___parse_many_small` | 2.768 ms | 2.829 ms | +2.2% | informational |
| `test__threaded___threadpool_parse_email` | 0.894 ms | 0.894 ms | +0.0% | informational |
| `test__threaded___threadpool_parse_email_small` | 14.268 ms | 13.770 ms | -3.5% | informational |

Noise floor from the pure-Python controls: **0.9%** (they cannot be affected by the build, so this is measurement error).

Worst treatment delta: **+15614.6%** (`test__fast_mail_parser___attachment_reread`).

</details>

## Checks run locally

`pytest tests --ignore=tests/benchmark` (807 passed, 2 skipped — the Linux-only `ru_maxrss` probes), `cargo fmt --all -- --check`, `cargo clippy --all-targets -- -D warnings -W clippy::cast_possible_truncation`, `mypy --strict fast_mail_parser/`, `ruff check .` — all green.

`vendor/`, `Cargo.toml` (no `py-clone`), `Cargo.lock` and `src/mail_parser.rs` are untouched — confirmed with `git diff --stat master -- vendor Cargo.toml Cargo.lock src/mail_parser.rs`, which is empty. The free-threading audit boundary is unaffected: the cell lives on the Python object in the binding layer, exactly like the lazy one the audit note already describes.

The M4 is not a proxy for x86 — the **Benchmark quality gate** job on this PR is the authoritative confirmation.

## Docs

`Readme.md` — the identity guarantee now holds in every mode, so the note next to `PyLazyAttachment` says that instead of "`PyAttachment.content` is exactly what it was". `__init__.pyi` gains one sentence each on `PyAttachment` and `PyMimePart`; signatures unchanged. `CHANGELOG.md` `[Unreleased]` entry with the numbers and the machine.


kurok:
Rebased onto `c1da4f8` now that #242 has landed. Only `CHANGELOG.md` conflicted (both PRs add an entry at the top of `[Unreleased]`); both entries are kept, #226's first since it is already on master.

Re-measured against the new base, since the old numbers were against `73f56da` and a stale base is its own bug:

| Benchmark | master `c1da4f8` | this branch | |
|---|---|---|---|
| `attachment_reread` | 12.8 µs | **0.083 µs** | ~156x |
| `full_read` | 0.257 ms | **0.231 ms** | −11.1% |
| `parse_many` | 1.987 ms | **1.874 ms** | −6.0% |
| `parse_message` | 0.230 ms | 0.226 ms | −1.6% (noise floor 1.8%) |

`full_read` reads −11.1% here against −2.9% in the original table. Same change, different base: #226 removed the `str` copy that used to sit in front of this path, so what is left is a larger share of a smaller number. The benchmark gate's own figure against the merge-base is the one to trust.

Local surface re-run after the rebase: 808 passed / 3 skipped, fmt, clippy, mypy --strict, ruff — all green.

PR #244, opened 2026-09-17
Decode base64 bodies with SIMD, keeping data-encoding as the arbiter (#228)

### LINKED ISSUE #228: Decode base64 with base64-simd in the vendored mailparse, keeping data-encoding as the rejection oracle
## Problem

After #213/#214/#218 the full parse of `tests/data/large_message.eml` (767 KiB, ~10,000 base64 lines of 76 chars) is 0.225 ms on an Apple M4 and 0.604 ms on the CI gate's EPYC, and roughly 80% of that is the base64 path: **53%** in `data_encoding`'s scalar decode loop (`decode_base_mut`) and **27%** in the whitespace strip that feeds it. The strip is already word-at-a-time and exact (`vendor/mailparse/src/bytescan.rs`); the remaining 53% is a 4-table-lookups-per-3-bytes scalar loop that is at scalar peak on both CPUs (~1.5 B/cycle on the M4, ~0.8 B/cycle on Zen 3). The 0.9.0 CHANGELOG entry says it in one line: "The two loops that made it 0-96% are fixed; what remains is the base64 decode proper."

The obvious fusion -- tell `data_encoding` to IGNORE all ASCII whitespace so the strip disappears -- does not work: `data_encoding` implements IGNORE as per-block error recovery (a `decode_wrap_block` detour every time the fast run hits an ignored byte), i.e. one detour per 76-char line, ~10,000 detours against the ~61 us the strip costs today. So the only lever left on the decode proper is wider lanes. `data_encoding`'s `decode` also allocates and zeroes a second large buffer (`vec![0u8; decode_len]`, 586 KB for this fixture) on top of the `cleaned` copy.

This issue proposes decoding `cleaned` with `base64-simd` (AVX2/SSE4.1 with runtime detection on x86-64, NEON on aarch64), in place, and falling back to today's `data_encoding::BASE64_MIME_PERMISSIVE` decode whenever the SIMD decoder rejects the input -- so every accepted message decodes to the same bytes and every rejected message produces the identical `MailParseError::Base64DecodeError(DecodeError { position, kind })`.

## Evidence

- `vendor/mailparse/src/body.rs:139-142`:
  ```rust
  fn decode_base64(body: &[u8]) -> Result<Vec<u8>, MailParseError> {
      let cleaned = crate::bytescan::strip_ascii_whitespace(body);
      Ok(data_encoding::BASE64_MIME_PERMISSIVE.decode(&cleaned)?)
  }
  ```
  Two passes, two large allocations, scalar decode. This is the only body-decode call site (`body.rs:29` installs it as the `Body::Base64` decoder); the binding decodes each part exactly once through `get_body_raw` (`src/mail_parser.rs:743`, `:846`, `:1375`) -- the #55 double decode is gone, so this function *is* the decode cost.
- `vendor/mailparse/src/bytescan.rs:65-90` -- `pub(crate) fn strip_ascii_whitespace(body: &[u8]) -> Vec<u8> { let mut cleaned = Vec::with_capacity(body.len()); ...` -- the 27% pass; word-at-a-time, exact, tested against the naive filter at every alignment (`strip_matches_filter`, `bytescan.rs:124-157`). Keep it.
- `data-encoding 2.6.0` (pinned at `Cargo.lock:27-31`; source under `~/.cargo/registry/src/*/data-encoding-2.6.0/src/lib.rs`):
  - `lib.rs:421-438` `decode_block`: `let y = values[input[j] as usize]; check!(j, y < 1 << bit); x |= u64::from(y) << ...` -- one table lookup per input symbol, u64 shift assembly, per 4-symbol block. `lib.rs:443-458` `decode_mut` loops it; `lib.rs:603-616` `decode_base_mut` is the wrapper that shows up at 53%.
  - `lib.rs:1494-1498` `pub fn decode(...) { let mut output = vec![0u8; self.decode_len(input.len())?]; ... output.truncate(len); }` -- the second allocation, zero-filled.
  - `lib.rs:744-790` `decode_wrap_mut` / `lib.rs:706-742` `decode_wrap_block`: every IGNORE byte ends the fast `decode_pad_mut` run with a `DecodePartial` and routes one block through a `[0u8; 8]` staging buffer with `skip_ignore` -- why an inline-ignore `Specification` would regress, not fuse.
  - `lib.rs:2429-2447` doc of `BASE64_MIME_PERMISSIVE`: `spec.padding = Some('=')`, wrap `\r\n` (IGNORE), `check_trailing_bits = false`. Padding is required; non-zero trailing bits are accepted; `=` at a block boundary mid-stream is accepted (see `BASE64.decode(b"SGVsbA==byB3b3JsZA==") == b"Hello world"` at `lib.rs:1485`).
- `vendor/mailparse/src/lib.rs:40` `Base64DecodeError(data_encoding::DecodeError)` and `:86-90` `impl From<data_encoding::DecodeError> for MailParseError` -- the public error type carries data-encoding's error; `src/fast_mail_parser.rs:67-72` maps it to Python `DecodeError` with `format!("Message parsing error: {error}")`. Keeping data-encoding as the rejection path preserves this text exactly.
- `vendor/mailparse/src/lib.rs:1869-1871` `test_base64_content_encoding_multiple_strings` parses `aGVsbG 8gd\r\n29ybGQ=\r\nZm9vCg==` -- mid-stream padding that a strict decoder rejects and this crate must keep accepting. It is the existing oracle for the fallback path.
- `vendor/mailparse/PATCH.md:7-8` ("`memchr = \"2.7.0\"` is added to this crate's `[dependencies]`") and `:49-63` (dependency-free where no slower; `memchr` kept for the byte search because every dependency-free variant measured +9-14%) -- the precedent and the bar a second dependency has to clear. `6302493` (#217) made this directory a permanent carry, so a dependency added here is a local decision.
- `Cargo.lock:6-9` `base64 0.22.1` (via `charset`) is scalar too -- nothing already in the graph offers SIMD.
- `deny.toml:22-38` allowlist includes MIT; `deny.toml:52-55` crates.io only. `Cargo.toml:5` `rust-version = "1.83"`; `rust-toolchain.toml` pins 1.98.0.
- `base64-simd 0.8.0` (verified on docs.rs, 2026-08-30): `pub fn decode_inplace<'d>(&self, data: &'d mut [u8]) -> Result<&'d mut [u8], Error>`; `STANDARD = Base64 { kind: Kind::Standard, extra: Extra::Pad }`; with `Extra::Pad` `decoded_length` enforces `ensure!(len % 4 == 0)`, `=` is counted only in the last two positions, and `decode_extra` rejects non-zero trailing bits (`(y2 & 0x0f) == 0`, `(y3 & 0x03) == 0` unless forgiving). Crate: MIT, edition 2021, `rust-version = "1.63"`, deps `vsimd 0.8.0` + `outref 0.5.0`, default features `std` + `detect` (runtime CPU detection). Last release 2022-12-28.
- `src/fast_mail_parser.rs:746` `pub content: Vec<u8>` on `PyAttachment`, copied to `PyBytes` on every `.content` access at `:768` -- the decoded `Vec` is retained for the Python object's lifetime, so its *capacity* matters (see Proposal).

## Proposal

In `vendor/mailparse/src/body.rs::decode_base64`: keep `strip_ascii_whitespace`, decode the whitespace-free buffer with `base64_simd::STANDARD`, fall back to `data_encoding::BASE64_MIME_PERMISSIVE` on any SIMD rejection. Because `STANDARD` accepts a strict subset of what `BASE64_MIME_PERMISSIVE` accepts on a whitespace-free buffer (same alphabet; padding required by both; SIMD additionally rejects mid-stream `=` and non-zero trailing bits), the set "SIMD accepts, data-encoding rejects" is empty, and both produce the same bytes on anything both accept. The fallback therefore makes acceptance, output and every error `position`/`kind` identical to today.

**Trap the first draft must not fall into:** `Base64::decode_inplace` is `decoded_length()` then `multiversion::decode::auto(src, dst, ...)` with `src == dst` -- it writes decoded bytes progressively over the input, so on `Err` the buffer is partially overwritten. The fallback must re-derive the stripped buffer from `body`, never reuse `cleaned`.

```rust
// vendor/mailparse/src/body.rs
fn decode_base64(body: &[u8]) -> Result<Vec<u8>, MailParseError> {
    let mut cleaned = crate::bytescan::strip_ascii_whitespace(body);
    // Strict RFC 4648 decode, SIMD where the CPU has it. Everything it accepts,
    // BASE64_MIME_PERMISSIVE accepts with the same bytes (same alphabet, padding
    // required by both; this one is additionally strict about mid-stream '=' and
    // non-zero trailing bits), so a success here is exactly what the previous
    // decoder would have produced.
    if let Ok(decoded) = base64_simd::STANDARD.decode_inplace(&mut cleaned) {
        let n = decoded.len();
        cleaned.truncate(n);
        return Ok(cleaned);
    }
    // Cold path: the decoder this crate has always used decides, so that every
    // rejection is the same MailParseError::Base64DecodeError { position, kind } as
    // before, and the two lenient cases (mid-stream padding, non-zero trailing
    // bits) still decode. `cleaned` is not reusable: decode_inplace writes over
    // its input as it goes.
    let cleaned = crate::bytescan::strip_ascii_whitespace(body);
    Ok(data_encoding::BASE64_MIME_PERMISSIVE.decode(&cleaned)?)
}
```

**Capacity.** In-place decode leaves `cleaned` with `capacity == body.len()` (~1.33x the decoded length) and `PyAttachment` retains that `Vec` for the object's lifetime (`src/fast_mail_parser.rs:746`). Today's output is exactly sized. Two ways to handle it; **measure both** and keep the faster one that does not regress memory unreasonably:
  (a) in-place as above plus `cleaned.shrink_to_fit()` (a `realloc` down; usually in place for large blocks but not guaranteed), or
  (b) out-of-place into an exactly-sized uninitialised buffer: `let mut out = Vec::with_capacity(base64_simd::STANDARD.decoded_length(&cleaned)?); STANDARD.decode(&cleaned, outref::Out::from_uninit_slice(out.spare_capacity_mut()))` then `unsafe { out.set_len(n) }` (or `decode_type::<Vec<u8>>` which does this internally). Still removes the 586 KB zero-fill; costs one allocation; keeps `cleaned` intact so the fallback needs no re-strip.
  Document the choice next to the function.

**New dependency** in `vendor/mailparse/Cargo.toml`: `base64-simd = "0.8"` (MIT; pulls `vsimd` and `outref`, both MIT -- `cargo deny check` confirms; MSRV 1.63 < this crate's 1.83; `default-features` on, for `detect`). Justification, to be recorded in `PATCH.md` next to `memchr`'s: the dependency-free decoder is at scalar peak, so unlike the strip there is no dependency-free variant that is "no slower"; only SIMD moves this number. Note honestly that base64-simd's last release is 2022-12 and it is `unsafe`-heavy SIMD; `cargo audit`/`cargo deny` are the gates, and RustSec shows no advisory at the time of writing.

**Behaviour that must stay identical:** every accepted input decodes to the same bytes; every rejected input raises the same Python `DecodeError` with the same message; `mode="metadata"` and `mode="lazy"` untouched-attachment paths do not change (they never call `decode_base64`); the RFC 2047 encoded-word decode in `vendor/mailparse/src/header.rs:48` stays on `data_encoding` (tiny inputs, not worth the surface).

**Expected impact** (not measured; the harness below confirms): decode proper 53% -> ~18-28% of today's time on the M4 (NEON, ~2-3x) and ~11-16% on the EPYC gate (AVX2, ~3-4x from a lower scalar base); one 586 KB zero-fill gone. Full parse roughly -25 to -35% (M4) and -35 to -45% (EPYC). Only x86-64 and aarch64 wheels get SIMD; `armv7`, `s390x`, `ppc64le`, `x86` (i686 gets SSE4.1/AVX2 via detection) fall back to base64-simd's scalar path -- check that path does not lose to data-encoding on the one non-SIMD target you can run (e.g. `RUSTFLAGS` cannot disable `detect`; use the `vsimd` fallback by building for a target without SIMD, or accept it unmeasured and say so in the PR).

1. `vendor/mailparse/Cargo.toml`: add `[dependencies.base64-simd] version = "0.8"`. Run `cargo update -p base64-simd` from the repo root so `Cargo.lock` gains `base64-simd`, `vsimd`, `outref`; do the same in `fuzz/` (it has its own lockfile and `[patch.crates-io]` at `fuzz/Cargo.toml:47-48`).
2. `vendor/mailparse/src/body.rs:139-142`: replace `decode_base64` as sketched. Do not touch `header.rs:48`.
3. `vendor/mailparse/src/body.rs`: add a `#[cfg(test)] mod tests` with a differential test `simd_and_data_encoding_agree` over a generated corpus in the style of `bytescan.rs:112-122` (`corpus()`): valid padded input at every length 0..=200, unpadded tails, `=` mid-stream at and off block boundaries, `A===`, `AA=A`, every non-alphabet byte at every position of a 4-block, both non-zero trailing-bit patterns (`QR==`, `QUI=`), every whitespace byte (`\t \n \x0c \r ' '`) interleaved, `\x0b` and `\x00`. Oracle: `data_encoding::BASE64_MIME_PERMISSIVE.decode(&naive_strip(body))` -- assert `Ok` bytes equal, or both `Err` with equal `DecodeError` (it derives `PartialEq`). This is what proves the "SIMD accepts, data-encoding rejects" set is empty.
4. `fuzz/`: add a target `base64_agreement` (`fuzz/Cargo.toml` `[[bin]]`, `data-encoding = "2.6.0"` as a fuzz dependency): wrap arbitrary bytes as `b"Content-Transfer-Encoding: base64\r\n\r\n" ++ data`, `mailparse::parse_mail(..).get_body_raw()`, compare with the oracle above (Ok bytes equal; Err Display equals `format!("Base64 decode error: {}", oracle_err)`). Give it the four-line `canary_armed` check every target carries (CONTRIBUTING.md "Fuzzing"), and add it to the `deep-fuzz.yml` matrix and the 60 s PR pass in `test.yml`.
5. Python pins for the lenient cases that now go through the fallback, in `tests/test_decode_errors.py`: `test_midstream_padding_still_decodes` (`aGVsbG8=` + `d29ybGQ=` concatenated as one body -> `b"helloworld"`), `test_nonzero_trailing_bits_still_decode` (`QR==` -> `b"A"`), `test_missing_padding_still_raises` (`QUJD` minus its pad -> `DecodeError`). Existing oracles that must stay green unchanged: `tests/test_decode_errors.py`, `tests/test_error_taxonomy.py` (`BROKEN_BASE64_BODY` -> `DecodeError`), `tests/test_stdlib_parity.py` (no new entry in `DIVERGENCES`), `tests/test_lazy_mode.py`, and the vendored suite (`test_base64_content_encoding`, `test_base64_content_encoding_multiple_strings` at `vendor/mailparse/src/lib.rs:1855-1880`).
6. `vendor/mailparse/PATCH.md`: update the header ("two functions changed ... and one dependency added" -> two dependencies), add the `decode_base64` paragraph describing the SIMD-then-oracle design and why a dependency is justified here when it was not for the strip (`:49-63`), extend the `diff -r` list (`:28-29`) and the "Keeping this in sync" step 2 (`:82-83`) to name `base64-simd`. The root `Cargo.toml:51-53` comment describes only the memmem change; refresh it in passing.
7. `CHANGELOG.md` `[Unreleased]` -> `### Changed`: one entry with the measured before/after from the harness below, in the style of the 0.9.0 entries.
8. Traps: never run cargo inside `vendor/mailparse` without `--target-dir` elsewhere -- `pyproject.toml:48` sweeps `vendor/mailparse/**/*` into the sdist and a `target/` there would ship. The vendored crate is a `[patch]` path dependency, not a workspace member: root `cargo clippy` lints path dependencies by default (hence its `--no-deps` flag) so the new code must be clippy-clean under `-D warnings -W clippy::cast_possible_truncation`, but root `cargo fmt --all` does not format it -- run `cargo fmt --manifest-path vendor/mailparse/Cargo.toml -- --check` yourself. No Python API change: `__init__.pyi`, `tests/test_contract.py` frozen sets and `docs/compatibility.md` stay untouched. abi3 wheels need no `target-cpu`: `detect` does runtime dispatch.

## Harness -- how to prove it

Build both sides once, then interleave measurements (the pattern `.github/workflows/test.yml:395-455` uses; a one-shot A/B cannot resolve 7%, see `ab_median.py` docstring):

```bash
# baseline = master 73f56da in a worktree, candidate = this branch; each in its own venv
git worktree add /tmp/fmp-base 73f56da
python -m venv /tmp/venv-base && CARGO_TARGET_DIR=/tmp/target-base /tmp/venv-base/bin/pip install -q "/tmp/fmp-base[test]"
python -m venv /tmp/venv-head && CARGO_TARGET_DIR=/tmp/target-head /tmp/venv-head/bin/pip install -q ".[test]"
# this revision's benchmark code measures BOTH builds
for r in 1 2 3; do for side in head base; do
  /tmp/venv-$side/bin/pytest -q tests/benchmark --benchmark-min-rounds=25 --benchmark-json=$side-$r.json
done; done
python .github/scripts/check_benchmark.py head-1.json   # absolute floor vs pure-Python mail-parser
```

Benchmarks in `tests/benchmark/test_read_message.py` expected to **improve together** (all decode every attachment): `test__fast_mail_parser___parse_message` (the gate; expected roughly -25% or better on the M4), `test__fast_mail_parser___parse_message_strict`, `test__fast_mail_parser___parse_many`, `test__fast_mail_parser___parse_tree`, `test__fast_mail_parser___parse_lazy_all_attachments`, `test__fast_mail_parser___full_read`. **Flat controls** (never call `decode_base64`; must stay within the noise floor the pure-Python rows show): `test__fast_mail_parser___parse_metadata`, `test__fast_mail_parser___parse_lazy_untouched`, `test__fast_mail_parser___parse_tree_metadata`, `test__fast_mail_parser___parse_tree_lazy_untouched`, `test__fast_mail_parser___parse_many_metadata`. No benchmark may regress more than 7%. Put the local before/after table (with the CPU line `ab_median.py` prints) in the PR body. **The M4 is not a proxy for x86**: the PR gate runs interleaved on the x86 runner and is the verdict for AVX2; if the gate's numbers disagree with the M4 by more than the control spread, re-run once before concluding (CONTRIBUTING.md "Performance"). Optionally re-sample a full parse the way the maintainer did and confirm `decode_base_mut` has left the top of the profile.

Run the two capacity variants ((a) in-place + `shrink_to_fit`, (b) out-of-place uninit) through the same harness and report both; pick per the numbers.

Correctness and lint surface, all must pass:

```bash
cargo fmt --all -- --check
cargo fmt --manifest-path vendor/mailparse/Cargo.toml -- --check
ruff check .
cargo deny check      # licences for base64-simd, vsimd, outref; crates.io source
cargo audit
cd fuzz && cargo +nightly fuzz run base64_agreement -- -max_total_time=60   # plus the existing two targets
```

## Acceptance criteria

- [ ] `vendor/mailparse/src/body.rs::decode_base64` decodes with `base64_simd::STANDARD` and falls back to `data_encoding::BASE64_MIME_PERMISSIVE` on rejection; the fallback never reuses a buffer `decode_inplace` has written into.
- [ ] New vendored differential test `simd_and_data_encoding_agree` passes and covers: unpadded tails, mid-stream `=` on and off block boundaries, non-zero trailing bits, every non-alphabet byte, every ASCII whitespace byte, `\x0b`/`\x00`.
- [ ] `vendor/mailparse/src/lib.rs` `test_base64_content_encoding_multiple_strings` (mid-stream padding) still passes unchanged.
- [ ] New fuzz target `base64_agreement` (with canary) is in `fuzz/Cargo.toml`, the PR fuzz step and the `deep-fuzz.yml` matrix, and runs 60 s clean.
- [ ] New Python tests `test_midstream_padding_still_decodes`, `test_nonzero_trailing_bits_still_decode`, `test_missing_padding_still_raises` pass; `tests/test_decode_errors.py`, `tests/test_error_taxonomy.py`, `tests/test_stdlib_parity.py` (no new `DIVERGENCES` entry), `tests/test_contract.py` pass unchanged.
- [ ] Interleaved local A/B (3 rounds) shows `test__fast_mail_parser___parse_message` faster by a margin that clears the pure-Python control spread, every decoding benchmark listed above improved, no benchmark regressed > 7%, and the metadata/lazy-untouched controls flat; numbers and CPU line are in the PR body.
- [ ] The PR benchmark gate (x86 runner) passes; its numbers are quoted in the CHANGELOG entry alongside the M4 numbers.
- [ ] `cargo deny check` and `cargo audit` pass with `base64-simd`, `vsimd`, `outref` in `Cargo.lock` (and `fuzz/Cargo.lock`).
- [ ] `PATCH.md` describes the second dependency, why it clears the bar the strip did not, the sync steps name it, and the `diff -r` list is accurate; root `Cargo.toml` `[patch.crates-io]` comment refreshed.
- [ ] `CHANGELOG.md` `[Unreleased]` has the entry; the capacity choice ((a) or (b)) is documented at the function and its memory consequence stated in the PR.
- [ ] Full lint/test surface above is green; no `target/` directory exists under `vendor/mailparse/`.

## Out of scope

- `base64_simd::forgiving_decode_inplace` / `remove_ascii_whitespace_inplace` (WHATWG forgiving mode accepts missing padding, which changes acceptance semantics) -- a separate, explicit decision.
- Replacing `data_encoding` entirely (it is the public `MailParseError::Base64DecodeError` payload) or touching the RFC 2047 encoded-word decode in `header.rs:48`.
- Quoted-printable decoding, charset conversion, `target-cpu`, PGO, custom allocators.
- Any change to the Python API, `__init__.pyi`, or `docs/compatibility.md`.

## Related issues

- #225
- #235 — regenerate `upstream.patch` / PATCH.md with any vendor change
- #240 — interpret a layout swing on this decode-path change with the salted A/B

kurok:
Closes #228.

Branched from `master` (`73f56da`), independent of #242 and #243.

## What changed

After #213/#214/#218 fixed the two byte-at-a-time loops, what remained of a full parse was the base64 decode proper: **53%** of it in `data_encoding`'s scalar loop — four table lookups per three bytes, at scalar peak on both CPUs tried. The 0.9.0 changelog said it in one line: *"what remains is the base64 decode proper."*

The only lever left on that is wider lanes. `decode_base64` now decodes the whitespace-free buffer with `base64_simd::STANDARD` (AVX2/SSE4.1 with runtime detection on x86-64, NEON on aarch64), and falls back to `data_encoding::BASE64_MIME_PERMISSIVE` on any rejection.

**The fallback is what makes this a speed change and not a change to what this library accepts.** `STANDARD` accepts a strict *subset* on a whitespace-free buffer — same alphabet, padding required by both, and `STANDARD` is additionally strict about `=` appearing mid-stream and about non-zero trailing bits, which are exactly the two lenient cases this library has always accepted. So "SIMD accepts, `data_encoding` rejects" is empty, anything both accept decodes to the same bytes, and every rejection still carries `data_encoding`'s own `DecodeError { position, kind }` into `MailParseError::Base64DecodeError` and reaches Python with identical text.

## How that claim is held up

Three independent things, because "the fast path is a subset" is an argument, not evidence:

1. **`simd_and_data_encoding_agree`** (vendored, `src/body.rs`) sweeps a built corpus against the old decoder as oracle: every length to 200 padded and unpadded, **every byte at every position** of a one- and two-block body, mid-stream `=` on and off block boundaries, both trailing-bit shapes, every ASCII whitespace byte interleaved, and `\x0b`/`\x00` — which are *not* whitespace, so they survive the strip and must be rejected. Asserts equal bytes on accept and an equal `DecodeError` on reject.
2. **`base64_agreement`**, a new fuzz target, asserts the same three-way agreement — acceptance, bytes, *and* error text — on arbitrary input through the public entry point. **3,578,785 executions clean in 60 s** locally. I also confirmed its canary still fires (`FMP_FUZZ_CANARY=1` → deliberate panic), since an unarmed canary is a silent hole.
3. **Three Python tests** pin the lenient cases that now reach the fallback (`test_midstream_padding_still_decodes`, `test_nonzero_trailing_bits_still_decode`, `test_missing_padding_still_raises`). Without them a fast path that quietly swallowed those would still look green.

## The shape of the call is not cosmetic

The issue asks for both capacity strategies to be measured. They were, and the result is the most interesting thing in this PR.

| form | what it does | full parse |
|---|---|---|
| `decode_inplace` over the stripped buffer, then `truncate` | reuses the strip's allocation, no second one | **9% slower** |
| `decode_to_vec` into an exactly-sized buffer *(shipped)* | one extra allocation, no zero-fill | **51% faster** |

Same machine, same corpus, same decoder — and **standalone the two are within 10% of each other** (77 µs vs 70 µs on a 781 KB payload). The in-place form is the one that reads better, which is precisely why the function and `PATCH.md` both carry a note saying not to "tidy" it back.

Getting there ruled things out in order rather than guessing: the fixture takes the fast path (all three parts are strict-clean — length % 4 = 0, no mid-stream `=`, zero trailing bits), so it wasn't silently falling back; `shrink_to_fit` wasn't the cost (identical 0.274 ms with and without); and **linking `base64-simd` without calling it** costs 1.5–3.4%, at or below that run's noise floor, so it isn't mere linkage either. What's left is the code-layout sensitivity of #204, which this crate builds squarely into: `lto = true`, `codegen-units = 1`.

## Numbers

Apple M4 (10 vCPU), rustc 1.98.0, CPython 3.12. **Two independent** interleaved A/Bs pooled to 8 rounds per side (`ab_median.py`). Pure-Python controls within **1.1%**.

|---|---|---|---|
| `parse_message` | 0.254 ms | **0.168 ms** | **−34%** |
| `parse_tree` | 0.246 ms | **0.159 ms** | −35% |
| `full_read` | 0.261 ms | **0.172 ms** | −34% |
| `parse_many` (8 × 767 KiB) | 1.989 ms | **1.318 ms** | −34% |

`mode="metadata"` (+0.2%) and `parse_lazy_untouched` (+1.5%) never call this function and are flat, which is the control that says the gain is the decode and not the weather.

<details><summary>Full pooled A/B report (8 rounds/side)</summary>

Median of 8 interleaved rounds per side; each value is a benchmark's minimum. Positive delta = `master` is slower.

| Benchmark | simd-to-vec | master | Delta | |
|---|---|---|---|---|
| `test__fast_mail_parser___full_read` | 0.172 ms | 0.261 ms | +51.9% |  |
| `test__fast_mail_parser___parse_lazy_all_attachments` | 0.176 ms | 0.265 ms | +50.8% |  |
| `test__fast_mail_parser___parse_lazy_untouched` | 0.043 ms | 0.043 ms | +1.5% |  |
| `test__fast_mail_parser___parse_many` | 1.318 ms | 1.989 ms | +50.9% |  |
| `test__fast_mail_parser___parse_many_metadata` | 0.240 ms | 0.238 ms | -0.7% |  |
| `test__fast_mail_parser___parse_message` | 0.168 ms | 0.254 ms | +50.7% |  |
| `test__fast_mail_parser___parse_message_strict` | 0.166 ms | 0.253 ms | +52.0% |  |
| `test__fast_mail_parser___parse_metadata` | 0.030 ms | 0.030 ms | +0.2% |  |
| `test__fast_mail_parser___parse_tree` | 0.159 ms | 0.246 ms | +55.0% |  |
| `test__fast_mail_parser___parse_tree_lazy_untouched` | 0.042 ms | 0.042 ms | -0.3% |  |
| `test__fast_mail_parser___parse_tree_metadata` | 0.031 ms | 0.031 ms | +0.4% |  |
| `test__mail_parser___parse_message` | 5.490 ms | 5.548 ms | +1.1% | control |
| `test__mailparser_lib___full_read` | 5.874 ms | 5.873 ms | -0.0% | control |
| `test__stdlib_email___full_read` | 7.471 ms | 7.483 ms | +0.2% | control |
| `test__threaded___parse_many` | 0.734 ms | 1.015 ms | +38.2% | informational |
| `test__threaded___parse_many_metadata_small` | 2.538 ms | 2.550 ms | +0.5% | informational |
| `test__threaded___parse_many_small` | 2.762 ms | 2.746 ms | -0.5% | informational |
| `test__threaded___threadpool_parse_email` | 0.630 ms | 0.892 ms | +41.6% | informational |
| `test__threaded___threadpool_parse_email_small` | 14.079 ms | 14.182 ms | +0.7% | informational |

Noise floor from the pure-Python controls: **1.1%** (they cannot be affected by the build, so this is measurement error).

Worst treatment delta: **+55.0%** (`test__fast_mail_parser___parse_tree`).

</details>

## Supply chain

`base64-simd` 0.8.0 is MIT and pulls `vsimd` 0.8.0 and `outref` 0.5.2, both MIT, all from crates.io — covered by `deny.toml`'s allowlist and source policy. I verified the licenses by hand locally; `cargo deny` and `cargo audit` are not installed on this machine, so **the CI jobs are the actual gate** for them.

Stated plainly rather than buried: base64-simd's last release is **2022-12** and it is `unsafe`-heavy SIMD. `PATCH.md` says so, and notes that the fallback bounds the blast radius of a *rejection* bug to a slow path but bounds nothing about a *wrong-bytes* bug — which is what the differential test and the fuzz target exist for.

## Checks run locally

`pytest tests --ignore=tests/benchmark` (806 passed, 2 skipped), the vendored suite (`cargo test --manifest-path vendor/mailparse/Cargo.toml --target-dir` elsewhere — 53 + 25 passed, including `test_base64_content_encoding_multiple_strings` unchanged), `cargo fmt --all --check` **and** `cargo fmt --manifest-path vendor/mailparse/Cargo.toml --check` (the root one does not format the vendored crate), `cargo clippy --all-targets -- -D warnings -W clippy::cast_possible_truncation`, `mypy --strict`, `ruff check .` — all green. No `target/` under `vendor/mailparse/`.

The M4 is not a proxy for x86, and the issue expects a *larger* win on the gate's EPYC (AVX2 from a lower scalar base). The **Benchmark quality gate** job is the authoritative number; I'll add its result to the CHANGELOG entry once it reports.

## Housekeeping

The PR fuzz budget stays at a minute — three targets at 20 s rather than two at 30 s — and `deep-fuzz.yml` gains the target in its matrix. `.gitignore` gains what `cargo fuzz` leaves behind, since running the new target locally is now something a contributor will do.


kurok:
Rebased onto `c1da4f8` now that #242 has landed. `CHANGELOG.md` was the only conflict; both entries kept.

**Worth recording why this PR showed only CodeQL results before the rebase.** It was opened after #242 merged, so it conflicted with master from the moment it was created. GitHub evaluates `pull_request`-triggered workflows against the PR's *merge commit* — and a conflicting PR has none, so `test.yml` never ran at all. The four CodeQL checks passed and nothing else existed, which is a state that reads as "green" at a glance while Lint, the test matrix, the benchmark gate, the fuzz pass and `cargo deny`/`cargo audit` had not run. The rebase makes the PR mergeable and the full suite is running now.

Re-measured against the new base:

| Benchmark | master `c1da4f8` | this branch | |
|---|---|---|---|
| `parse_message` | 0.239 ms | **0.157 ms** | **−52%** |
| `parse_tree` | 0.248 ms | **0.157 ms** | −57% |
| `full_read` | 0.257 ms | **0.172 ms** | −50% |
| `parse_many` (8 × 767 KiB) | 1.989 ms | **1.326 ms** | −50% |

Controls within 0.3%. One honest note: `parse_many_metadata` read **+2.8% slower** in this run, above that run's noise floor. Metadata mode never calls `decode_base64`, so it is either layout or the measurement; it is well inside the 7% gate, and I will re-check it against the gate's own numbers rather than assert it away.

kurok:
Rebased onto `abb31b8` (post-#243). Green: 15/15, `CLEAN`.

**One thing the rebase turned up that is worth recording, because it is the #204 failure mode again and it nearly read as a blocker.**

Re-measuring on the M4 against the new base — i.e. with #227 in master — the decode paths were still ~−40%, but the paths that *never call* `decode_base64` came back **13–19% slower**:

| Benchmark | M4 vs `abb31b8` | x86 gate (this run) |
|---|---|---|
| `parse_metadata` | **−19.1%** (slower) | −1.5% (faster) |
| `parse_metadata_str` | **−18.8%** (slower) | +2.1% |
| `parse_many_metadata` | **−18.7%** (slower) | −0.8% |
| `parse_lazy_untouched` | **−12.6%** (slower) | −4.2% (faster) |
| `parse_tree_metadata` | **−18.3%** (slower) | −0.1% |

That is above the 7% threshold, so I did not assume it away. What I established before pushing:

- It is **not** anything the code does at runtime. With base64-simd **linked but never executed** (`if black_box(false) { … }`, work otherwise identical to master) the metadata paths are still 14–19% slower.
- It is deterministic, not noise: per-round minima have zero spread — `0.030 0.030 0.030` on master against `0.037 0.037 0.037` on the branch, controls at 0.7%.
- `#[inline(never)]` on `decode_base64` — the remedy this repo's issues prescribe for exactly this — did not move it. It is not on the branch; the diff stays minimal.
- It did **not** exist when this branch was based on `c1da4f8`. It emerges only in combination with #227, which is what makes it a placement effect rather than a property of either change.

The gate above is the verdict, and it is flat: worst delta +2.1%, inside the threshold, controls at 0.5%. So this is an M4-only artifact — the mirror image of #204, where Zen showed swings the M4 never did.

Recording it because it is a live example of the thing `PATCH.md` warns about, and because the next person to rebase this branch onto a different master may well see it again and should not spend the afternoon I did rediscovering that it is placement.

PR #245, opened 2026-09-17
Decode quoted-printable bodies a run at a time (#229)

### LINKED ISSUE #229: Vendored mailparse: replace quoted_printable::decode with a run-copying QP decoder, and benchmark a quoted-printable message
## Problem

Quoted-printable is the transfer encoding of most `text/html` newsletters and transactional mail, and it is the last transfer decoder in this library that still runs byte-at-a-time. The two loops fixed in 0.9.0 (MIME boundary search, base64 whitespace strip -- see `vendor/mailparse/PATCH.md`) were each "a test and a bounds-checked push per byte" and together cost 0-96% of a parse depending on where the linker placed them. The quoted-printable path has the same shape, three times over: the `quoted_printable` 0.5.1 crate that the vendored mailparse delegates to (a) copies the whole body through a `filter_map(..).collect::<String>()`, (b) walks it again with `str::lines()` + `trim_end()`, and (c) decodes with a `bytes.next()` / `decoded.push(byte)` loop into a `Vec::new()` that was never given a capacity. Our own `quoted_printable_invalid_escape` in `src/mail_parser.rs` then makes a fourth pass over the same bytes.

Nothing in the benchmark suite can see this. Every benchmark in `tests/benchmark/test_read_message.py` runs over `tests/data/large_message.eml` (8bit + three base64 parts) or a synthetic message with no `Content-Transfer-Encoding`, so the maintainer's profile of a full parse (53% `data_encoding::decode_base_mut`, 27% whitespace strip, 7.5% memmem) describes base64 mail only. A QP-heavy message goes through none of the fast paths built so far, and the CI gate would not notice if it got slower.

This issue replaces the body-decoding call in the vendored crate with a run-copying decoder (memchr between `=` and `\n` events, `extend_from_slice` for everything in between, one allocation sized to the input) that returns byte-identical output to `quoted_printable::decode(.., ParseMode::Robust)`, and adds the benchmark that makes the change measurable and gated.

## Evidence

- `vendor/mailparse/src/body.rs:144-149` -- the entire QP body path is the external crate:
  ```rust
  fn decode_quoted_printable(body: &[u8]) -> Result<Vec<u8>, MailParseError> {
      Ok(quoted_printable::decode(
          body,
          quoted_printable::ParseMode::Robust,
      )?)
  }
  ```
  Contrast `body.rs:140-143`, the already-fixed base64 path: `let cleaned = crate::bytescan::strip_ascii_whitespace(body); Ok(data_encoding::BASE64_MIME_PERMISSIVE.decode(&cleaned)?)`. The dispatch is at `body.rs:33-37` (`"quoted-printable" => Body::QuotedPrintable(EncodedBody { decoder: decode_quoted_printable, .. })`).
- `Cargo.lock:167-168` pins `quoted_printable` 0.5.1 (`vendor/mailparse/Cargo.toml:49-50` requires `0.5.0`). In that crate's `src/lib.rs:202-300` (`~/.cargo/registry/src/*/quoted_printable-0.5.1/`), `_decode` does:
  - `:202-209` `let filtered = input.into_iter().filter_map(|&c| match c { b'\t' | b'\r' | b'\n' | b' '..=b'~' => Some(c as char), _ => None }).collect::<lib::String>();` -- a char-by-char copy of the whole body into a `String` that starts empty (filter_map's lower size hint is 0, so it regrows);
  - `:213` `let mut decoded = lib::Vec::new();` -- no capacity, so ~15 regrowths for a 90 KB body;
  - `:214-216` `let mut lines = filtered.lines(); ... Some(v) => v.trim_end().bytes()` -- second pass;
  - `:234-286` the inner `loop { let byte = match bytes.next() ... if byte == b'=' { ... } else { decoded.push(byte); } }` -- third pass, one bounds-checked push per byte;
  - `:291-298` `if filtered.ends_with('\n') { decoded.push(b'\r'); decoded.push(b'\n'); }`.
- `src/mail_parser.rs:104-132` -- `quoted_printable_invalid_escape`: `while at < raw.len() { if raw[at] != b'=' { at += 1; continue; } ... }` over `body.get_raw()`. Called for every part in the full parse (`:1379`) and in lazy mode (`:818`), so a QP body is walked four times before it reaches Python.
- `vendor/mailparse/src/header.rs:52-63` -- the other user of the crate, RFC 2047 `Q` encoded-words (`input.replace('_', " ")`, `trim_end`, restore trailing whitespace). Inputs are header-sized; this stays on the crate.
- No benchmark exercises QP: `tests/data/large_message.eml:40` is `8bit`, `:57`, `:149`, `:3950` are `base64`; `tests/benchmark/test_read_message.py` uses only `large_message` and `_small_message()` (no CTE header).
- A fixture with real QP already exists: `tests/data/valid_message.eml:73` (text/plain, 8,122 bytes at lines 75-235) and `:238` (text/html, 89,932 bytes at lines 240-2282; 706 soft breaks, 2,905 `=XX` escapes). The file has zero CR bytes and zero 8-bit bytes on the wire, so every CRLF in the decoded bodies comes from the decoder's canonicalisation. `tests/conftest.py:38-39` already exposes it as the `valid_message` fixture.
- The Robust-mode quirks are documented contracts, not accidents, so the replacement must reproduce them exactly:
  - bare LF decodes to CRLF: `docs/compatibility.md:85-99` (section 4), `fast_mail_parser/__init__.pyi:183-186` ("quoted-printable emits a line break as CRLF, so a body of bare LFs decodes larger than it was encoded"), `tests/test_stdlib_parity.py:89-99` (DIVERGENCES for `invalid_message` text parts);
  - an invalid escape passes through as literal text and is reported, not repaired: `tests/test_parse_warnings.py:544-600` (`=ZZ` survives in content; `=3D` and `=\r\n` are not lossy; bare LF is not lossy);
  - vendored test `vendor/mailparse/src/lib.rs:1840-1851` `test_quoted_printable_content_encoding`.
- Maintenance claims that become stale: `vendor/mailparse/PATCH.md:3-5` ("two functions changed (one via a new module)"), `:27-28` ("`diff -r` against the registry copy shows exactly `src/bytescan.rs`, `src/lib.rs` ..., `src/body.rs` (one function), `Cargo.toml` and this file"), `:78-85` ("Keeping this in sync" step 2 lists the files to keep); `Cargo.toml:50-52` at the root still says "one function changed".
- Gate mechanics: `.github/workflows/test.yml:427-431` -- "This revision's benchmark code measures BOTH builds", so a benchmark added in the PR is compared against the base build in the same PR as long as it needs no new API; `:460` `AB_THRESHOLD_PCT: "7"`; `.github/scripts/check_benchmark.py:62-63` selects the absolute-floor pair by exact name (`test__fast_mail_parser___parse_message`, `test__mail_parser___parse_message`), so a new name cannot disturb it.

Known profile numbers for scale (Apple M4 / CI EPYC, maintainer-measured): full parse of the 767 KiB fixture 0.225 ms / 0.604 ms; metadata mode 0.030 ms. A 90 KB QP body decoded at the crate's ~3 ns/byte (three passes plus regrowth) is **expected** to cost 250-350 us on the M4 -- more than the entire large-message parse. That number is an estimate; the first implementation step below measures it.

## Proposal

New module `vendor/mailparse/src/qp.rs` with `pub(crate) fn decode_robust(input: &[u8]) -> Vec<u8>`, wired in at `body.rs:144-149` as `Ok(crate::qp::decode_robust(body))`. No new dependency: `memchr` is already a dependency of the vendored crate (`vendor/mailparse/Cargo.toml:40-41`) and `quoted_printable` stays for `header.rs:57`. `Cargo.lock` is unchanged.

Behaviour that must stay byte-identical to `quoted_printable::decode(input, ParseMode::Robust)` (read from the 0.5.1 source; the differential test is the arbiter, this list is the spec):

1. **Pre-filter.** Only bytes in `{0x09, 0x0A, 0x0D} ∪ [0x20, 0x7E]` take part. Everything else (`0x00-0x08`, `0x0B`, `0x0C`, `0x0E-0x1F`, `0x7F`, `>= 0x80`) is dropped *before* line splitting and trimming, so `b"abc=\x80\n"` is a soft break and `b"abc \x80\n"` trims to `abc`.
2. **Lines** are split on LF; a CR immediately before the LF belongs to the terminator (`str::lines` semantics: `"a\rb"` is one line, `"a\r\r\n"` yields `"a\r"`). A final line without LF is still a line.
3. **Per line**, trailing bytes in `{b' ', b'\t', b'\r'}` are trimmed (after the pre-filter these are the only whitespace chars `str::trim_end` can see). Do not use `<[u8]>::trim_ascii_end` -- its set differs.
4. **Escapes**, on the trimmed line: `=` + two hex digits (either case) -> one byte; `=` as the last byte -> soft break, nothing emitted and *no CRLF before the next line*; `=X` as the last two bytes -> literal `=X` and the line ends normally (CRLF follows); `=XY` with a non-hex X or Y -> the three literal bytes, decoding continues after Y (so `==3D` decodes to `==3D`).
5. **Line joins.** CRLF is emitted between consecutive lines except after a soft break. Empty lines produce a CRLF each.
6. **Trailing newline.** If the (filtered) input ends with LF, one CRLF is appended after the loop -- even when the last line ended in a soft break (`b"abc=\n"` -> `b"abc\r\n"`). Empty input -> empty output.

Sketch:

```rust
// vendor/mailparse/src/qp.rs
use memchr::memchr;

pub(crate) fn decode_robust(input: &[u8]) -> Vec<u8> {
    // Rule 1: the crate filters before it splits lines. Do the same, but only
    // when there is something to filter; `dirty` is a word-at-a-time scan using
    // bytescan's below_mask(x, 0x20) plus a `>= 0x7F` mask, then a run-copying
    // strip (same shape as strip_ascii_whitespace). Common case: no copy.
    let filtered;
    let src: &[u8] = if has_dirty_byte(input) { filtered = strip_dirty(input); &filtered } else { input };

    let mut out = Vec::with_capacity(src.len() + 2);
    let mut pos = 0;
    let mut pending_crlf = false; // the crate's `add_line_break == Some(true)`
    while pos < src.len() {
        let nl = memchr(b'\n', &src[pos..]).map(|i| pos + i);
        let (line_end, next) = match nl { Some(i) => (i, i + 1), None => (src.len(), src.len()) };
        let mut line = &src[pos..line_end];
        if let [rest @ .., b'\r'] = line { line = rest; }          // rule 2
        while let [rest @ .., b' ' | b'\t' | b'\r'] = line { line = rest; } // rule 3
        if pending_crlf { out.extend_from_slice(b"\r\n"); pending_crlf = false; }
        pending_crlf = decode_line(line, &mut out);                // rules 4-5; false only after a soft break
        pos = next;
    }
    if src.last() == Some(&b'\n') { out.extend_from_slice(b"\r\n"); } // rule 6
    out
}

/// Copies runs between `=` with memchr; per-byte work happens only at an escape.
fn decode_line(line: &[u8], out: &mut Vec<u8>) -> bool { /* ... returns `pending_crlf` */ }
```

Keep the per-byte code confined to the escape branch: that is where the layout sensitivity PATCH.md describes could come back, and a 90 KB HTML body has ~3,600 escape/line events against ~90,000 bytes.

1. **Measure first.** Add the benchmark from the Harness section (`test__fast_mail_parser___parse_qp_message`) and run it on master. Record the number in the PR body. If the QP decode turns out to be a small fraction of the `valid_message` parse (sample with `py-spy record --native` or `samply` on a loop of `parse_email(valid_message)`), stop and report -- the vendored surface is not worth it for a small win.
2. Create `vendor/mailparse/src/qp.rs` as above; add `mod qp;` next to `mod bytescan;` in `vendor/mailparse/src/lib.rs`. Make `bytescan::below_mask`, `word`, `first_hit`, `WORD`, `HI` `pub(crate)` if `qp.rs` reuses them (they are private `fn`s at `bytescan.rs:34-52`). The vendored crate is `edition = "2018"` and the root pins `rust-version = "1.83"` (`Cargo.toml:5`); `bytescan.rs:22-24` already notes `as_chunks` is off-limits -- stay on APIs stable at 1.83. No `unsafe`.
3. Wire `body.rs:144-149`: `fn decode_quoted_printable(body: &[u8]) -> Result<Vec<u8>, MailParseError> { Ok(crate::qp::decode_robust(body)) }`. Keep the signature (it is stored as a `fn` pointer at `body.rs:34`) and keep `MailParseError::QuotedPrintableDecodeError` (`lib.rs:37`, `:80-84`) -- it is public API of the crate and `header.rs` still produces it.
4. **Differential test in the vendored crate** (`#[cfg(test)] mod tests` in `qp.rs`, modelled on `bytescan.rs:91-150`): for every input, `assert_eq!(decode_robust(&h), quoted_printable::decode(&h, quoted_printable::ParseMode::Robust).unwrap())`. Corpus: the xorshift generator from `bytescan.rs:104-120` at every length 0..=96 with an alphabet that hits every rule -- `b"=", b"3", b"D", b"a", b"f", b"Z", b" ", b"\t", b"\r", b"\n", b"\x0b", b"\x0c", b"\x7f", b"\x00", b"\xff"`; plus, for every pair `(x, y)` in `0..=255`, the inputs `[b'=', x, y]`, `[b'=', x, y, b'\n']`, `[b'a', b'=', x]`, `[b'=', x]` and `[b'=', x, b'\n']`; plus hand-written cases for rules 1-6 (`b"abc=\n"`, `b"a=\nb\n"`, `b"\n"`, `b""`, `b"a\r\r\n"`, `b"abc \x80\n"`, `b"==3D"`, `b"x=  \r\n"`).
5. **Fuzz target** `fuzz/fuzz_targets/qp_agreement.rs`: add `quoted_printable = "0.5.1"` to `fuzz/Cargo.toml` `[dependencies]` (the fuzz crate is excluded from the sdist/wheel by `pyproject.toml:52`, so this is a dev-only dependency; keep the version equal to `Cargo.lock:168`). The target builds `b"Content-Transfer-Encoding: quoted-printable\r\n\r\n" ++ data`, calls `mailparse::parse_mail(..).unwrap().get_body_raw().unwrap()` and asserts byte equality with `quoted_printable::decode(data, ParseMode::Robust).unwrap()`. Register it as a `[[bin]]` like the two existing ones (`fuzz/Cargo.toml:26-38`) and add `qp_agreement` to the matrix at `.github/workflows/deep-fuzz.yml:68`. Run it locally for at least 10 minutes before opening the PR (`cargo +nightly fuzz run qp_agreement -- -max_total_time=600`).
6. **Docs that must change with `vendor/`:** `vendor/mailparse/PATCH.md` -- "two functions changed" (`:3-5`) becomes three, list `src/qp.rs` in the `diff -r` claim (`:27-28`) and in "Keeping this in sync" step 2 (`:80-81`), and add a third numbered entry under "The changes" with the before/after numbers; root `Cargo.toml:50-52` comment (already stale) and the `.github/workflows/test.yml:57-58` comment. `CHANGELOG.md` gets an entry under `## [Unreleased]` / `### Changed` with the measured numbers and the CPU they were measured on.
7. **Traps.** Never run `cargo` inside `vendor/mailparse` without `--target-dir` outside the tree: `pyproject.toml:46-50` sweeps `vendor/mailparse/**/*` into the sdist and a `target/` there ships. No API moves, so `fast_mail_parser/__init__.pyi` and the frozen sets in `tests/test_contract.py` stay untouched -- if the diff touches either, something went wrong. `src/mail_parser.rs` is compiled by path into the fuzz crate (`fuzz/Cargo.toml:6-12`); this issue does not change it. The base64 path next door is not to be touched: replacing `strip_ascii_whitespace` with `BASE64_MIME_PERMISSIVE`'s `ignore` would make data-encoding fall back to its scalar per-line path at every CRLF.
8. The M4 is not a proxy for x86: the placement effect PATCH.md documents is a Zen phenomenon. The PR gate runs on the CI's EPYC and is the x86 verdict; if it reads a regression on any existing benchmark, try `#[inline(never)]` on `decode_robust`/`decode_line` and re-measure before anything else.

## Harness -- how to prove it

**New benchmark** in `tests/benchmark/test_read_message.py` (name is free-form; the gate pair is selected by exact name, see the module docstring and `check_benchmark.py:62-63`):

```python
def test__fast_mail_parser___parse_qp_message(valid_message: str, benchmark: Callable):
    # The only fixture whose bodies are quoted-printable: 89,932 bytes of text/html
    # and 8,122 of text/plain, bare-LF on the wire. Every other benchmark here is
    # base64/8bit, so this is the one that sees the QP decoder at all.
    from fast_mail_parser import parse_email

    payload = valid_message.encode()
    mail = parse_email(payload)
    assert mail.text_html and mail.text_plain, "both QP parts must decode"
    assert "=3D" not in mail.text_html[0], "the HTML part must actually be decoded"
    assert "\r\n" in mail.text_html[0], "robust decoding canonicalises bare LF to CRLF"

    benchmark(parse_email, payload)
```

Optionally a `test__fast_mail_parser___parse_qp_message_metadata` twin (guarded with the `TypeError` skip pattern used by `test__fast_mail_parser___parse_metadata`) as the non-decoding control on the same fixture.

**Local interleaved A/B** (the arbiter; never compare against a stored JSON from another day):

```sh
# two venvs, one wheel each; never `maturin develop` into the source tree
uv venv /tmp/venv-master && uv venv /tmp/venv-pr
for v in /tmp/venv-master /tmp/venv-pr; do uv pip install --python $v/bin/python maturin pytest pytest-benchmark mail-parser; done
git worktree add /tmp/fmp-master master
(cd /tmp/fmp-master && /tmp/venv-master/bin/maturin build --release -o /tmp/wheels-master) && uv pip install --python /tmp/venv-master/bin/python /tmp/wheels-master/*.whl
/tmp/venv-pr/bin/maturin build --release -o /tmp/wheels-pr && uv pip install --python /tmp/venv-pr/bin/python /tmp/wheels-pr/*.whl
# the PR checkout's tests/benchmark measures both builds, exactly as test.yml:427-431 does
~/.cache/fmp-bench/ab_local.sh "$PWD" /tmp/venv-master master /tmp/venv-pr pr 3
```

`ab_local.sh` runs `.github/scripts/ab_median.py --informational test__threaded___` over three interleaved rounds (`--benchmark-min-rounds=25`). Read it as follows:

- `test__fast_mail_parser___parse_qp_message`: expected to improve. The decoder itself is expected 4-8x faster and the end-to-end parse of `valid_message` (headers included) 1.5-2.5x; the PR body states the measured median on both sides. An improvement under 20% means the vendored surface is not paying for itself -- report and close.
- Every existing benchmark (`test__fast_mail_parser___parse_message`, `_strict`, `_full_read`, `_parse_many`, `_parse_tree*`, `_parse_metadata`, `_parse_lazy_*`, `_parse_many_metadata`): expected unchanged (the fixture has no QP part); none may regress more than 7% (`AB_THRESHOLD_PCT`), and the pure-Python controls (`test__mail_parser___`, `test__mailparser_lib___`, `test__stdlib_email___`) must stay inside their noise floor for the treatment delta to count.
- The CI gate repeats this on x86 (EPYC); its verdict, not the M4's, decides the layout question.

**Full lint/test surface** (all must pass):

```sh
cargo fmt --all -- --check
ruff check .
cargo +nightly fuzz run qp_agreement -- -max_total_time=600   # from fuzz/
```

Correctness oracles inside `pytest tests`: `tests/test_stdlib_parity.py` (the `invalid_message`/`valid_message` DIVERGENCES at `:89-99` are the CRLF contract), `tests/test_parse_warnings.py:544-600` (Robust pass-through and the warnings channel), `tests/test_lazy_mode.py` (lazy decode equals full decode), `tests/test_known_issues.py` (`invalid_message.eml`).

## Acceptance criteria

- [ ] `vendor/mailparse/src/body.rs::decode_quoted_printable` no longer calls `quoted_printable::decode`; `vendor/mailparse/src/header.rs:57` still does.
- [ ] `vendor/mailparse/src/qp.rs` exists, uses only `std` and the already-present `memchr`, contains no `unsafe`, and compiles on the pinned toolchain with `rust-version = "1.83"` semantics.
- [ ] `Cargo.lock` is unchanged (no new dependency in the shipped graph); `fuzz/Cargo.toml` gains only `quoted_printable = "0.5.1"`.
- [ ] The in-crate differential test compares `decode_robust` against `quoted_printable::decode(.., Robust)` over the generated corpus at every alignment, every `=xy` byte pair, and the hand-written rule cases, and passes under `cargo test --manifest-path vendor/mailparse/Cargo.toml --target-dir /tmp/vendor-target`.
- [ ] `fuzz/fuzz_targets/qp_agreement.rs` exists, is registered in `fuzz/Cargo.toml` and in the `deep-fuzz.yml` matrix, and ran locally for at least 10 minutes with zero findings (say so in the PR body).
- [ ] `test__fast_mail_parser___parse_qp_message` exists in `tests/benchmark/test_read_message.py`, asserts a decoded HTML body once outside the timed call, and improves in the interleaved A/B (measured number, both CPUs, in the PR body).
- [ ] No existing benchmark regresses more than 7% in the interleaved A/B locally or in the CI gate.
- [ ] `pytest tests --ignore=tests/benchmark`, `cargo fmt --check`, `cargo clippy --all-targets -D warnings -W clippy::cast_possible_truncation`, `mypy --strict fast_mail_parser/`, `ruff check .` all pass.
- [ ] `vendor/mailparse/PATCH.md` describes three changed functions, lists `src/qp.rs` in the `diff -r` claim and the sync checklist, and records the before/after numbers; the root `Cargo.toml` and `test.yml` comments about the vendored copy are brought up to date.
- [ ] `CHANGELOG.md` `[Unreleased]` has an entry with the measured numbers and the CPU.
- [ ] `fast_mail_parser/__init__.pyi`, `tests/test_contract.py` and `docs/compatibility.md` are untouched (no API or behaviour change).

## Out of scope

- `src/mail_parser.rs:104-132` (`quoted_printable_invalid_escape`): after this change it becomes the largest remaining per-byte pass over a QP body. Removing it means the vendored decoder returning the first invalid-escape offset (a `pub` addition to the vendored crate's API) or a word-at-a-time `=` search in the core. Decide with a profile sample of the new benchmark after this lands; file separately.
- The RFC 2047 `Q` path in `header.rs` (header-sized inputs, different semantics: `_` -> space, trailing-whitespace restore).
- Any change to the base64 path or to `strip_ascii_whitespace`.
- Upstreaming: `vendor/mailparse` is a permanent carry (#217); the QP crate is a separate project and is not touched.

## Related issues

- #223
- #230 — removes the fourth pass (`quoted_printable_invalid_escape`) around this decoder
- #235
- #240

kurok:
Closes #229.

Branched from `master` (`abb31b8`). Independent of #244, but both touch `vendor/mailparse/PATCH.md` and the fuzz target list, so whichever lands second will want a small rebase — noted at the bottom.

## What changed

Quoted-printable is the transfer encoding of most HTML newsletters and transactional mail, and it was the last transfer decoder here still running byte-at-a-time. `decode_quoted_printable` handed the whole body to the `quoted_printable` crate, which makes **three passes** over it: a char-by-char copy into a `String` through a `filter_map`, a second walk with `lines()` + `trim_end()`, then a decode loop with one bounds-checked `push` per byte into a `Vec` that was never given a capacity.

Bodies now go through `vendor/mailparse/src/qp.rs`: one allocation sized to the input, `memchr` to find line breaks and escapes, `extend_from_slice` for everything between them — so a line with no `=` in it is a single copy. Plain `std` plus the `memchr` already vendored here. No `unsafe`, no new dependency, `Cargo.lock` unchanged. The RFC 2047 encoded-word path in `header.rs` still uses the crate: header-sized inputs, different semantics.

## The benchmark had to come first

**Nothing in the suite could see this path.** Every fixture in `tests/benchmark/` is base64 or 8bit — which is also why the profile everyone quotes (53% base64 decode, 27% whitespace strip) describes base64 mail only. A quoted-printable message goes through none of those fast paths, and the gate would not have noticed this getting *slower*.

So per the issue's step 1, I added the benchmark and measured on master before writing a line of the decoder: **218 µs**, against **22 µs** for metadata mode on the same message. ~90% of that parse was the QP decode. That is what justified touching the vendored surface at all.

## Numbers

Apple M4 (10 vCPU), rustc 1.98.0, CPython 3.12, 5 interleaved rounds (`ab_median.py`). Pure-Python controls within **3.2%**.

|---|---|---|---|
| `parse_qp_message` *(new)* | 0.219 ms | **0.134 ms** | **−39%** |
| `parse_qp_message_metadata` *(new control)* | 0.022 ms | 0.022 ms | flat |

The control is the point: same fixture, same headers, same structure, no transfer decode. It does not move, so the gain is the decode and not the parse around it. Every other benchmark landed inside the noise floor.

**One honest caveat on the size of the win.** The issue estimated the decoder itself at 4–8× and end-to-end at 1.5–2.5×. End-to-end came in at 1.64×, inside that range, but the decode component went 197 µs → 112 µs, i.e. ~1.76× rather than 4–8×. The likely reason is named in the issue's own Out of scope: `quoted_printable_invalid_escape` in `src/mail_parser.rs` still makes a full per-byte pass over every QP body, and it is now the largest remaining one. That is a separate issue, and this PR does not touch it.

| Benchmark | qp | master | Delta | |
|---|---|---|---|---|
| `test__fast_mail_parser___full_read` | 0.242 ms | 0.241 ms | -0.5% |  |
| `test__fast_mail_parser___parse_lazy_all_attachments` | 0.254 ms | 0.251 ms | -1.5% |  |
| `test__fast_mail_parser___parse_lazy_untouched` | 0.043 ms | 0.043 ms | +0.4% |  |
| `test__fast_mail_parser___parse_many` | 1.875 ms | 1.873 ms | -0.1% |  |
| `test__fast_mail_parser___parse_many_metadata` | 0.243 ms | 0.239 ms | -1.8% |  |
| `test__fast_mail_parser___parse_message` | 0.229 ms | 0.231 ms | +1.2% |  |
| `test__fast_mail_parser___parse_message_strict` | 0.231 ms | 0.233 ms | +0.8% |  |
| `test__fast_mail_parser___parse_metadata` | 0.030 ms | 0.030 ms | +0.0% |  |
| `test__fast_mail_parser___parse_metadata_str` | 0.030 ms | 0.030 ms | +0.3% |  |
| `test__fast_mail_parser___parse_qp_message` | 0.134 ms | 0.219 ms | +63.0% |  |
| `test__fast_mail_parser___parse_qp_message_metadata` | 0.022 ms | 0.022 ms | +0.2% |  |
| `test__fast_mail_parser___parse_tree` | 0.234 ms | 0.231 ms | -1.5% |  |
| `test__fast_mail_parser___parse_tree_lazy_untouched` | 0.042 ms | 0.042 ms | +0.1% |  |
| `test__fast_mail_parser___parse_tree_metadata` | 0.031 ms | 0.031 ms | +0.3% |  |
| `test__mail_parser___parse_message` | 5.592 ms | 5.414 ms | -3.2% | control |
| `test__mailparser_lib___full_read` | 5.935 ms | 5.875 ms | -1.0% | control |
| `test__stdlib_email___full_read` | 7.561 ms | 7.563 ms | +0.0% | control |
| `test__threaded___parse_many` | 0.993 ms | 0.955 ms | -3.8% | informational |
| `test__threaded___parse_many_metadata_small` | 2.546 ms | 2.574 ms | +1.1% | informational |
| `test__threaded___parse_many_small` | 2.771 ms | 2.806 ms | +1.3% | informational |
| `test__threaded___threadpool_parse_email` | 0.860 ms | 0.861 ms | +0.1% | informational |
| `test__threaded___threadpool_parse_email_small` | 14.126 ms | 14.210 ms | +0.6% | informational |

</details>

## Byte-identical, asserted three ways

The entire safety argument is that output does not change, so it is tested rather than claimed:

1. **In-crate differential test** (`qp.rs`), against `quoted_printable::decode(.., Robust)` as oracle: a generated corpus at every length to 96 over an alphabet that reaches every rule; **every `(x, y)` pair after `=`** in the four positions where the crate's escape handling branches (`=x`, `=x\n`, `a=x`, `=xy`, `=xy\n` — 65,536 pairs); hand-written cases for each of the six rules (soft break, literal `=X` at end of line, non-hex octet resuming after `Y`, the trimmed set, bare CR, dropped bytes, empty lines, trailing newline); and the real fixture.
2. **`qp_agreement` fuzz target**, same equality on arbitrary input through the public entry point: **22,934,494 executions in 601 s, zero findings.** Canary included and verified to fire.
3. The vendored suite's existing `test_quoted_printable_content_encoding`, and the Python `test_parse_warnings.py` / `test_stdlib_parity.py` contracts, all unchanged and green.

## Why `quoted_printable` is now pinned to `=0.5.1`

Writing that differential test is what found this, and it is a latent bug independent of this PR.

**0.5.1 and 0.5.2 do not agree.** For a body whose last line ends in a soft break, `b"abc=\n"` decodes to `b"abc\r\n"` on 0.5.1 and `b"abc"` on 0.5.2 — `if filtered.ends_with('\n')` gained `&& add_line_break == Some(true)`. The requirement here was `0.5.0`, a caret, so:

- `cargo update` could have changed body **and encoded-word** decoding, with nothing in the suite to catch it; and
- this crate's own `cargo test` resolves its **own** lockfile, independent of the root one — so it was pulling 0.5.2 while the extension links 0.5.1. My first test run failed for exactly that reason, and a differential test that compares against a version you do not ship is worse than no test.

The pin closes both. `Cargo.lock` is unchanged: the root already resolved to 0.5.1. `src/qp.rs` reproduces 0.5.1, because that is what this library ships.

## Checks run locally

`pytest tests --ignore=tests/benchmark` (808 passed, 3 skipped), the vendored suite (55 + 25 passed) via `--target-dir` outside the tree, `cargo fmt --all --check` **and** `cargo fmt --manifest-path vendor/mailparse/Cargo.toml --check`, `cargo clippy --all-targets -- -D warnings -W clippy::cast_possible_truncation`, `mypy --strict`, `ruff check .` — all green. No `target/` under `vendor/mailparse/`. `__init__.pyi`, `tests/test_contract.py` and `docs/compatibility.md` are untouched, as the issue requires — there is no API or behaviour change.

The M4 is not a proxy for x86; the **Benchmark quality gate** is the verdict.

## Rebase note

#244 (base64 SIMD) also edits `vendor/mailparse/PATCH.md`, the root `Cargo.toml` `[patch.crates-io]` comment, `fuzz/Cargo.toml` and the `FUZZ_TARGETS` list — each adding a third fuzz target and a third changed function. Whichever merges second needs those merged by hand rather than taken wholesale; the conflicts are additive and small. #244 also adds `.gitignore` entries for what `cargo fuzz` leaves behind, which I deliberately did not duplicate here.


kurok:
Rebased onto `e7bbb05` now that #244 has landed. Five files conflicted, all additively — `PATCH.md`, the root `Cargo.toml` patch comment, `fuzz/Cargo.toml`, `test.yml` and `deep-fuzz.yml` each gained a third changed function / fourth fuzz target. Both sides kept; PR fuzz budget stays at a minute (four targets × 15 s).

**Two things the rebase turned up that are worth flagging.**

**1. #244 left a small gap in `PATCH.md` that I fixed here.** Its "The changes" entry 2 still described only the whitespace strip — the paragraph about `base64_simd::STANDARD` and the `data_encoding` fallback never made it into master, and the `diff -r` claim still said `src/body.rs (one function)`. Since "the `diff -r` list is accurate" is an acceptance criterion for this issue, entry 2 now describes both halves and the list names `src/qp.rs` and the `mod tests`.

**2. The metadata benchmarks read ~23% *faster* on this branch, and that is not this PR's doing.** Re-measured on the M4 against the new base:

| Benchmark | master `e7bbb05` | this branch | |
|---|---|---|---|
| `parse_qp_message` | 0.221 ms | **0.137 ms** | **+61%** — the actual change |
| `parse_metadata` | 0.037 ms | 0.030 ms | +23% — see below |
| `parse_many_metadata` | 0.295 ms | 0.241 ms | +22% |
| `parse_tree_metadata` | 0.038 ms | 0.031 ms | +23% |
| `parse_lazy_untouched` | 0.050 ms | 0.043 ms | +16% |
| `parse_qp_message_metadata` | 0.023 ms | 0.022 ms | +3.4% (control, flat) |

Those metadata numbers are the M4 code-placement artifact I documented on #244 — master sits at 0.037 ms there, where it was 0.030 ms before base64-simd was linked, and this change happens to shift the layout back. It is not a quoted-printable win: metadata mode never transfer-decodes. The x86 gate measured those same benchmarks as flat on #244, so I expect it to show them flat here too, and **the gate is the number to quote, not this table**. Reporting it because a 23% swing on an unrelated benchmark should be explained rather than pocketed.

The claim this PR actually makes is the first row and its control: same fixture, same headers, decode removed — 0.221 → 0.137 ms.

Re-verified after the rebase: 811 passed / 3 skipped, the vendored suite 57 + 25 (all four QP differential tests among them), fmt for both manifests, clippy, mypy --strict, ruff.

PR #246, opened 2026-09-17
Cache the headers dict per object, collapse the six getters, and freeze the result classes (#231)

### LINKED ISSUE #231: Cache the `headers` dict per object, collapse the six identical getters, and mark the result classes `frozen`
## Problem

Every result class that exposes `headers` -- `PyMail`, `PyMailMetadata`, `PyLazyMail`, `PyMimePart`, `PyMimePartMetadata`, `PyLazyMimePart` -- stores the headers as a Rust `Vec<(String, Vec<String>)>` and rebuilds a brand-new Python `dict` on **every** attribute read: one `PyDict`, one `str` per key, one `list` per key, one `str` per value. For `tests/data/valid_message.eml` (29 keys, 32 fields) that is ~91 Python object allocations per read; the README's own example reads `email.headers` twice per message. Python callers reasonably treat `.headers` as a stored attribute and do `m.headers.get("From")`, `m.headers.get("Subject")`, `"X" in m.headers` ... each of which is a full rebuild. The parse-time marshalling in this binding is deliberately minimal (fields stay Rust `String`/`Vec`, one `Py::new` per result), so in `mode="metadata"` -- the header-sweep mode, 0.030 ms per large message on the M4, ~2 us per 0.8 KB message in the 2000-message batch -- a few header reads per message are expected to cost as much as the parse itself. None of this shows in any existing benchmark, because none reads `headers` inside the timed call.

The same six-line getter body is copy-pasted six times, so any change to header semantics (as #157 was) has to be made in six places. And none of the eleven `#[pyclass]`es is `frozen` even though not one has a `&mut self` method or a `#[pyo3(set)]` field, and the module docs already claim full immutability; PyO3 therefore runs an atomic compare-exchange borrow flag on every getter call for nothing.

The crate already contains the exact fix pattern: `PyLazyAttachment.content` is a `OnceLock<Py<PyBytes>>` published once so that `a.content is a.content`. `headers` should be the same.

## Evidence

All lines from master `73f56da`.

- `src/fast_mail_parser.rs:265-271` (PyMailMetadata), `:360-366` (PyMimePart), `:457-463` (PyMimePartMetadata), `:525-531` (PyLazyMimePart), `:863-869` (PyMail), `:1105-1111` (PyLazyMail) -- six byte-identical bodies:
  ```rust
  #[getter]
  fn headers<'py>(&self, py: Python<'py>) -> PyResult<Bound<'py, PyDict>> {
      let dict = PyDict::new(py);
      for (name, values) in &self.headers {
          dict.set_item(name, values)?;
      }
      Ok(dict)
  }
  ```
- `src/fast_mail_parser.rs:258, 338, 435, 497, 826, 1092` -- each class stores `pub headers: Vec<(String, Vec<String>)>`; no cache field exists. Constructor sites that fill it: `:319, :407, :615, :654, :902, :1160`.
- `src/fast_mail_parser.rs:858-861` (PyMail doc comment): "Built on access rather than stored as a dict ... the previous `HashMap` field, converted per access, produced a different order every time (#157)" -- #157 fixed order, not the per-access rebuild.
- `src/fast_mail_parser.rs:963, 984-987, 1054` -- `PyLazyAttachment`: `content: OnceLock<Py<PyBytes>>`, getter `if let Some(cached) = self.content.get() { return Ok(cached.bind(py).clone()); }`, publish via `self.content.get_or_init(|| bytes).bind(py).clone()`. `:517` -- `PyLazyMimePart` has the same field. The publish-once pattern and its race rationale (`:1029-1037`) are already in the crate.
- `#[pyclass]` attributes at `:196, 240, 332, 431, 493, 671, 719, 741, 788, 925, 1065` -- eleven classes, none `frozen`. `grep -n '&mut self\|pyo3(set)' src/fast_mail_parser.rs` returns nothing. `src/mail_parser.rs:9-26` states the thread-safety invariant: "no shared mutable state ... Nothing here is shared and nothing here is mutable"; the binding layer's only interior mutability is the `OnceLock` cache.
- `~/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/pyo3-0.29.2/src/pycell/impl_.rs:66-90, 150-157` -- `BorrowFlag::increment` is an `AcqRel` `compare_exchange` loop plus a `fetch_sub` on release, run by `try_borrow` for every `&self` getter on a non-frozen class; frozen classes use the no-op `EmptySlot` checker.
- `Readme.md:806` `for name, values in email.headers.items():` and `Readme.md:812` `for hop in email.headers.get("Received", []):` -- two full rebuilds per message in the documented idiom. `tests/test_mime_tree.py:157` has the same `"Subject" in part.headers` then `part.headers["Subject"]` shape.
- `Readme.md:163` -- `2000 x 0.8 KB, mode="metadata"` = 4.20 ms for the batch, i.e. ~2 us per message; a 4-header small message's dict is ~13 allocations, so one extra `headers` read per message is expected to be a large fraction of that.
- `tests/test_headers.py:5` `assert len(valid_mail.headers) == 29`; the fixture has 32 header fields (`awk` over the header block), so ~91 objects per read. `tests/data/large_message.eml` has 24 fields (<=73 objects).
- Known profile (maintainer, M4): full parse 0.225 ms, metadata 0.030 ms; hot loops are base64 decode 53%, whitespace strip 27%, memmem 7.5%. **Nothing in this issue touches those loops**; the gate benchmarks are expected to be flat.
- **Documented behaviour that changes:** `docs/migrating.md:253-254` -- "**Header mutation.** `headers` is a plain dict snapshot; changing it changes nothing." With a cached dict, mutating the returned dict becomes visible on later `headers` reads of the *same object* (the parse result -- `subject`, `from_`, a fresh `parse_email` -- is still unaffected). No test pins the snapshot semantics: `tests/test_contract.py:131-137` only requires `isinstance(valid_mail.headers, dict)` with `str` keys and `list[str]` values; `tests/test_metadata_mode.py:49` and `tests/test_lazy_mode.py:117` compare by `==`. No test mutates `headers` (`grep -rn 'headers\[.*\] *=' tests/` is empty).

## Proposal

New dependency: **none**.

1. One shared type in `src/fast_mail_parser.rs` replacing the bare `Vec` field on all six classes:

   ```rust
   /// Header pairs in wire order plus their one-time Python projection.
   ///
   /// `OnceLock<Py<PyDict>>` for the same reason `PyLazyAttachment.content` is:
   /// every read returns the *same* dict, so the second read allocates nothing.
   pub(crate) struct Headers {
       pairs: Vec<(String, Vec<String>)>,
       dict: OnceLock<Py<PyDict>>,
   }

   impl Headers {
       pub(crate) fn new(pairs: Vec<(String, Vec<String>)>) -> Self {
           Headers { pairs, dict: OnceLock::new() }
       }

       /// Build on first access (never in a constructor, so the parse path is untouched).
       pub(crate) fn to_dict<'py>(&self, py: Python<'py>) -> PyResult<Bound<'py, PyDict>> {
           if let Some(cached) = self.dict.get() {
               return Ok(cached.bind(py).clone());
           }
           let dict = PyDict::new(py);
           for (name, values) in &self.pairs {
               dict.set_item(name, values)?;
           }
           // Fallible work stays outside `get_or_init` (no `get_or_try_init` on stable);
           // a racing loser's dict is dropped under the GIL we hold. Same trade as
           // `PyLazyAttachment::decode`.
           Ok(self.dict.get_or_init(|| dict.unbind()).bind(py).clone())
       }
   }
   ```

   Each of the six getters becomes `self.headers.to_dict(py)`; the six constructor sites wrap with `Headers::new(...)`. The field must **not** get `#[pyo3(get)]` (the public attribute set is frozen by `tests/test_contract.py`).

2. Add `frozen` to all eleven `#[pyclass]` attributes (`#[pyclass(frozen, skip_from_py_object)]`, and `#[pyclass(frozen)]` for `PyMail` at `:788`). All fields are `Sync` (`String`, `Vec<u8>`, `Option<_>`, `Vec<Py<_>>`, `OnceLock<Py<_>>`), there are no `&mut self` methods and no setters, so this compiles as-is. Land it as a **separate commit** (or PR) from step 1 so the CI A/B can attribute any code-layout swing.

3. Behaviour that must stay identical: `headers` is still a real `dict[str, list[str]]` (not a mappingproxy -- `tests/test_contract.py:131` needs `dict`), same keys in wire order, same values, equal across modes (`tests/test_metadata_mode.py:49`, `tests/test_lazy_mode.py:117`, `tests/test_tree_modes.py`), same for `str` and `bytes` payloads (`tests/test_contract.py:200, 221`, `tests/test_payload_memory.py:92`). New, to be pinned: `m.headers is m.headers` on every class; the dict is built lazily (a parse that never reads `headers` builds no dict).

4. Docs: rewrite `docs/migrating.md:253-254` to say `headers` is the same dict object on every read and should be treated as read-only -- mutating it does not change the parse (`subject`, `from_`, re-parsing) but is visible on later reads of that object. Update the `headers` docstrings in `fast_mail_parser/__init__.pyi` (six classes: lines ~46-47, 98, 148, 231, 398, 434-437) and the PyMail attribute table wording at `Readme.md:210` if it mentions building. Rust doc comments at `:858-861` ("Built on access rather than stored as a dict") need the same update.

1. Read `src/fast_mail_parser.rs:940-1056` first (PyLazyAttachment's cache, its doc comments and the race rationale); mirror its wording for `Headers`.
2. Introduce `Headers` near the top of `src/fast_mail_parser.rs` (after the `use` block, `:25` already imports `std::sync::OnceLock`). Replace the six `pub headers: Vec<(String, Vec<String>)>` fields (`:258, 338, 435, 497, 826, 1092`) with `pub(crate) headers: Headers`, the six getter bodies (`:265-271, 360-366, 457-463, 525-531, 863-869, 1105-1111`) with `self.headers.to_dict(py)`, and the six constructor sites (`:319, 407, 615, 654, 902, 1160`) with `Headers::new(...)`. None of the six classes derives `Clone` (verified: only PyAttachmentMetadata/ParseWarning/PyAddress/PyAttachment do), so `OnceLock` in the struct is fine. The fuzz targets (`fuzz/fuzz_targets/*.rs`) read `mail.headers` from the PyO3-free `mail_parser` core, not from these classes -- no change there.
3. Do **not** touch `src/mail_parser.rs`: its module doc (`:9-26`) forbids interior mutability in the core and explicitly places caches "on the Python object ... in the binding layer". Keep it that way.
4. Tests (oracle): `pytest tests --ignore=tests/benchmark` must stay green unchanged. Add to `tests/test_headers.py`:
   - `test__headers_is_the_same_dict_on_every_read`: parametrised over `parse_email(payload)`, `mode="metadata"`, `mode="lazy"`, and the root plus first child of `parse_email_tree(payload, mode=m)` for all three modes; asserts `obj.headers is obj.headers`.
   - `test__headers_is_built_on_first_read_and_shared_across_threads`: 8 threads each read `mail.headers` once on a fresh object; `len({id(d) for d in results}) == 1` (same shape as the `content` identity-across-threads test in `tests/test_lazy_mode.py`).
   - `test__mutating_the_headers_dict_does_not_change_the_parse`: `h = mail.headers; h["X-Injected"] = ["1"]; h.pop("Subject", None)`; assert `mail.subject` unchanged, `parse_email(payload).headers == original_copy`, and (pinning the new documented semantics) `mail.headers is h`.
   - Keep `tests/test_contract.py` attribute sets untouched -- the new field has no `#[pyo3(get)]`, so `_public_attrs` (`tests/test_contract.py:98-99`) must not change. `tests/test_readme_snippets.py` checks the README against the API; keep any README edit consistent.
5. `frozen` commit: change the eleven attributes, run `cargo clippy --all-targets -- -D warnings -W clippy::cast_possible_truncation`; if PyO3 rejects anything it is because a method takes `&mut self` or a field has `#[pyo3(set)]` -- there are none at 73f56da.
6. CHANGELOG: under `## [Unreleased]` add a `### Changed` entry stating (a) `headers` now returns the same dict object on every read, built on first access; (b) the observable consequence -- mutations of that dict persist across reads of the same object, the parse itself is unaffected, and `docs/migrating.md` was updated; (c) all result classes are `frozen`. If the maintainer considers the old "snapshot" sentence a contract, add the `breaking-change` label and bump the minor version.
7. Traps: no `vendor/` change (so no `PATCH.md`, no vendored `cargo test`); abi3 build unaffected (`Py<PyDict>` and `OnceLock` are ABI-neutral); never run `cargo` inside `vendor/mailparse` without `--target-dir` elsewhere (sdist include glob sweeps `target/`); `mypy --strict fast_mail_parser/` and `ruff check .` must pass after the `.pyi` edits (`ruff` runs over `tests/` too).

## Harness -- how to prove it

Build both sides once, then measure interleaved (the pattern `.github/workflows/test.yml:380-475` uses; single build-and-measure cycles read ~16% apart on a no-op per `.github/scripts/ab_median.py:7-12`):

```bash
# base = master 73f56da in a worktree, head = the branch
git worktree add /tmp/fmp-base 73f56da
for side in head base; do
  uv venv /tmp/venv-$side && VIRTUAL_ENV=/tmp/venv-$side uv pip install maturin pytest pytest-benchmark mail-parser
done
(cd /tmp/fmp-base && VIRTUAL_ENV=/tmp/venv-base CARGO_TARGET_DIR=/tmp/target-base /tmp/venv-base/bin/maturin develop --release)
VIRTUAL_ENV=/tmp/venv-head /tmp/venv-head/bin/maturin develop --release
# Both sides run THIS revision's tests/benchmark (as CI does), interleaved A,B,A,B,A,B
for round in 1 2 3; do for side in head base; do
  /tmp/venv-$side/bin/pytest -q tests/benchmark --benchmark-min-rounds=25 --benchmark-json=$side-$round.json
done; done
python .github/scripts/check_benchmark.py head-1.json   # absolute floor vs pure-Python mail-parser
```

New benchmarks to add to `tests/benchmark/test_read_message.py` (names start with `test__fast_mail_parser___` so they are gated, not informational). Both use only API that exists on the base, so the A/B reports both sides:

- `test__fast_mail_parser___headers_first_read`: `benchmark(lambda: parse_email(payload, mode="metadata").headers)` on `large_message`. Expected: flat vs base within noise (one build either way, plus one `OnceLock` publish).
- `test__fast_mail_parser___headers_repeat_read`: parse once outside the timed call, then `benchmark(lambda: (m.headers.get("From"), m.headers.get("Subject"), m.headers.get("Received")))`. Expected: base ~3 x (dict build for 24 fields, ~3-6 us on M4); head ~3 dict lookups (< 0.5 us) -- an order-of-magnitude drop. Report the measured ratio in the PR body.
- Optional: the same pair on `_small_message()` (`tests/benchmark/test_read_message.py:215`) to show the fraction of a ~2 us metadata parse an extra read costs.

Gate: no existing benchmark may regress > 7% (`AB_THRESHOLD_PCT=7`); in particular `test__fast_mail_parser___parse_message`, `test__fast_mail_parser___parse_metadata`, `test__fast_mail_parser___parse_tree*`, `test__fast_mail_parser___parse_lazy*`, `test__fast_mail_parser___parse_many*` are expected flat -- nothing on their path changes except the `frozen` attribute. The pure-Python `test__mail_parser___`/`test__stdlib_email___` controls are the noise floor; a treatment delta is only believable above it.

**Layout caveat:** the `frozen` commit changes generated getter code under fat LTO / `codegen-units = 1`, and #204 showed x86 Zen swinging up to 96% on loop alignment while the M4 shows nothing. Local M4 numbers go in the PR body but are not a proxy for x86; the PR gate on CI (x86 EPYC) is the confirmation, and `toolchain-ab.yml` (`workflow_dispatch`, `.github/workflows/toolchain-ab.yml:29-41`) can be pointed at the branch if the gate reads anything odd. If the `frozen` commit alone moves a gate benchmark, drop it and land the cache + helper on their own.

Full local surface, all must pass: `pytest tests --ignore=tests/benchmark`, `cargo fmt --all -- --check`, `cargo clippy --all-targets -- -D warnings -W clippy::cast_possible_truncation`, `mypy --strict fast_mail_parser/`, `ruff check .`. `vendor/` is not touched, so the vendored `cargo test` is not required.

## Acceptance criteria

- [ ] One `Headers` type with one `to_dict` implementation; `grep -c 'PyDict::new(py)' src/fast_mail_parser.rs` drops from 6 header-getter sites to 1 for headers.
- [ ] All six `headers` getters return the same object on repeated reads (`obj.headers is obj.headers`) for `PyMail`, `PyMailMetadata`, `PyLazyMail`, `PyMimePart`, `PyMimePartMetadata`, `PyLazyMimePart`; pinned by the new tests in `tests/test_headers.py`.
- [ ] The dict is built on first access, not in any constructor (`from_mail`/`from_metadata`/`from_node`/`from_lazy` do not touch `Headers.dict`).
- [ ] Existing suite green: `pytest tests --ignore=tests/benchmark`; `tests/test_contract.py` attribute sets unchanged.
- [ ] All eleven `#[pyclass]` attributes carry `frozen`, in a separate commit; clippy clean.
- [ ] Two new benchmarks added as named above; `headers_repeat_read` improves by >= 5x on the M4 (expected ~10x), `headers_first_read` within noise.
- [ ] Interleaved 3-round A/B vs 73f56da: no benchmark regresses > 7%; controls quoted; numbers and CPU in the PR body; CI gate green on x86.
- [ ] `docs/migrating.md:253-254` rewritten; `headers` docstrings in `fast_mail_parser/__init__.pyi` and the Rust doc comments at `src/fast_mail_parser.rs:858-861` updated; `mypy --strict fast_mail_parser/`, `ruff check .`, `cargo fmt --all -- --check` pass.
- [ ] CHANGELOG `[Unreleased] / Changed` entry describing the identity semantics, the mutation visibility, and `frozen`.

## Out of scope

- Caching `date_parsed` (`src/fast_mail_parser.rs:278-284, 871+`): a single `datetime` construction, cheap and rarely re-read.
- `PyMail.attachments` being a `#[pyo3(get)] Vec<PyAttachment>` (`:822-823`) with `#[derive(Clone)]` on `PyAttachment` (`:742`): every read of `mail.attachments` clones every attachment's `content: Vec<u8>` into fresh objects, and `PyAttachment.content` (`:762-765`) copies again into `PyBytes`. A larger access-time cost than headers; deserves its own issue with its own benchmark.
- `PyMimePart.content` (`:373`) copying per read.
- Any change to `src/mail_parser.rs`, `vendor/mailparse`, the decode loops, or the release profile.
- Returning a read-only mapping type instead of `dict` (breaks `tests/test_contract.py:131`).

## Related issues

- #227
- #233 — the getter collapse is specified here with caching; land this first
- #238 — the Rust-side header cost below this dict

kurok:
Closes #231.

Branched from `master` (`e7bbb05`). Independent of #245.

Three commits, deliberately: the headers cache, then `frozen`, then tests/benchmarks/docs. The issue asks for `frozen` to be separable so a code-layout swing can be attributed to one change or the other — which, given this repo's history, turned out to be the right instinct.

## What changed

Six classes expose `headers` — `PyMail`, `PyMailMetadata`, `PyLazyMail`, `PyMimePart`, `PyMimePartMetadata`, `PyLazyMimePart` — and each rebuilt a whole Python `dict` on **every read**: one `PyDict`, a `str` per key, a `list` per key, a `str` per value. About ninety objects for `valid_message.eml`, on an attribute Python callers reasonably treat as stored. The README's own idiom reads it twice per message.

All six now share one `Headers` type holding the pairs plus a `OnceLock<Py<PyDict>>` — the same publish-once pattern as `PyLazyAttachment.content`. The dict is built on **first access and never in a constructor**, so a parse that never reads `headers` pays nothing. The six byte-identical getter bodies collapse to one, which is what made #157 a six-place change.

Separately, the eleven `#[pyclass]`es are now `frozen`. None has a `&mut self` method or a `#[pyo3(set)]` field, so PyO3 was running an `AcqRel` compare-exchange borrow flag on every getter call, plus a `fetch_sub` on release, to guard mutation that cannot happen.

## Numbers

Apple M4 (10 vCPU), rustc 1.98.0, CPython 3.12, 5 interleaved rounds. Pure-Python controls within **2.0%**.

|---|---|---|---|
| `headers_repeat_read` *(new)* | 2.416 µs | **0.073 µs** | **33x** |
| `headers_first_read` *(new control)* | 37.50 µs | 30.54 µs | see below |
| `attachment_reread` | 0.083 µs | **0.042 µs** | 2x, from `frozen` alone |

`headers_repeat_read` is the README idiom — three lookups on an already-parsed message. 33x, against the issue's ≥5x bar and ~10x estimate.

`attachment_reread` halving is the `frozen` commit on its own: that benchmark is three `a.content` reads on shared objects after #227, so with the dict rebuild gone what is left is almost entirely PyO3's borrow-flag atomics. It is a small absolute number but a clean read on what `frozen` buys.

**Two readings in that table I am *not* claiming as wins.** `headers_first_read` (+22.8%) and every metadata/lazy-untouched benchmark (+16% to +24%) are the M4 code-placement artifact I documented on #244 — master sits at 0.037 ms on `parse_metadata` where it was 0.030 ms before base64-simd was linked, and touching this file happens to shift the layout back. The x86 gate measured those same benchmarks flat on both #244 and #245. The issue predicts `headers_first_read` should be flat, and on x86 I expect it to be. **The gate is the number to quote for those rows; the 33x is the claim this PR makes.**

| Benchmark | headers-cache | master | Delta | |
|---|---|---|---|---|
| `test__fast_mail_parser___attachment_reread` | 0.000 ms | 0.000 ms | +97.5% |  |
| `test__fast_mail_parser___full_read` | 0.164 ms | 0.171 ms | +4.2% |  |
| `test__fast_mail_parser___headers_first_read` | 0.031 ms | 0.037 ms | +22.8% |  |
| `test__fast_mail_parser___headers_repeat_read` | 0.000 ms | 0.002 ms | +3195.4% |  |
| `test__fast_mail_parser___parse_lazy_all_attachments` | 0.176 ms | 0.181 ms | +2.8% |  |
| `test__fast_mail_parser___parse_lazy_untouched` | 0.043 ms | 0.050 ms | +16.2% |  |
| `test__fast_mail_parser___parse_many` | 1.315 ms | 1.348 ms | +2.5% |  |
| `test__fast_mail_parser___parse_many_metadata` | 0.239 ms | 0.293 ms | +22.8% |  |
| `test__fast_mail_parser___parse_message` | 0.158 ms | 0.164 ms | +3.9% |  |
| `test__fast_mail_parser___parse_message_strict` | 0.156 ms | 0.163 ms | +3.9% |  |
| `test__fast_mail_parser___parse_metadata` | 0.030 ms | 0.037 ms | +23.5% |  |
| `test__fast_mail_parser___parse_metadata_str` | 0.030 ms | 0.037 ms | +23.8% |  |
| `test__fast_mail_parser___parse_tree` | 0.157 ms | 0.164 ms | +4.4% |  |
| `test__fast_mail_parser___parse_tree_lazy_untouched` | 0.042 ms | 0.048 ms | +16.3% |  |
| `test__fast_mail_parser___parse_tree_metadata` | 0.031 ms | 0.038 ms | +21.9% |  |
| `test__mail_parser___parse_message` | 5.513 ms | 5.401 ms | -2.0% | control |
| `test__mailparser_lib___full_read` | 5.809 ms | 5.765 ms | -0.8% | control |
| `test__stdlib_email___full_read` | 7.560 ms | 7.441 ms | -1.6% | control |
| `test__threaded___parse_many` | 0.745 ms | 0.712 ms | -4.5% | informational |
| `test__threaded___parse_many_metadata_small` | 2.577 ms | 2.573 ms | -0.1% | informational |
| `test__threaded___parse_many_small` | 2.803 ms | 2.813 ms | +0.4% | informational |
| `test__threaded___threadpool_parse_email` | 0.631 ms | 0.644 ms | +2.1% | informational |
| `test__threaded___threadpool_parse_email_small` | 14.463 ms | 14.198 ms | -1.8% | informational |

</details>

## The behaviour change, and why it is documented rather than hidden

`mail.headers is mail.headers` is now `True`, so **edits to the returned dict persist across later reads of that object**. The parse behind it does not move: `subject`, `from_` and the rest are unaffected, and re-parsing the payload gives the original headers back.

`docs/migrating.md` previously said "`headers` is a plain dict snapshot; changing it changes nothing". That sentence is now false in its second half, so it is rewritten to say to treat `headers` as read-only and to copy it (`dict(mail.headers)`) if you need one you can edit. The stub docstrings and the Rust doc comment say the same.

No test pinned the old snapshot semantics (`grep -rn 'headers\[.*\] *=' tests/` was empty), and `tests/test_contract.py` only requires `isinstance(headers, dict)` with `str` keys and `list[str]` values — all still true. **If you consider the old sentence a contract rather than an incidental property, this wants the `breaking-change` label and a minor bump; say so and I will add it.**

## Tests

Added to `tests/test_headers.py`, and I ran them against master first — the first three **fail** there and pass here, which is what makes them worth having:

- `test__headers_is_the_same_dict_on_every_read` — all six classes, via full/metadata/lazy and the tree root plus first child.
- `test__headers_is_built_on_first_read_and_shared_across_threads` — eight barrier-synchronised first readers, one `id()`.
- `test__mutating_the_headers_dict_does_not_change_the_parse` — pins the new semantics in both directions.
- `test__headers_is_still_a_plain_dict_of_str_to_list` and `test__headers_agree_across_modes` — invariants that pass either way.

`tests/test_contract.py` attribute sets are untouched: the new field carries no `#[pyo3(get)]`.

## Checks run locally

`pytest tests --ignore=tests/benchmark` (818 passed, 3 skipped), `cargo fmt --all --check`, `cargo clippy --all-targets -- -D warnings -W clippy::cast_possible_truncation`, `mypy --strict fast_mail_parser/`, `ruff check .` — all green. Clippy earned its keep here: it rejected a `pairs()` accessor I had added speculatively and nothing used.

No `vendor/` change, so no `PATCH.md` and no vendored `cargo test`. `src/mail_parser.rs` untouched — its module doc forbids interior mutability in the core and places caches "on the Python object ... in the binding layer", which is where this one is.


kurok:
**Gate status, and why I am not patching it here.**

This PR was green at 15/15 before I rebased it onto `2aee3df`. The only thing that changed is the base — #229's quoted-printable decoder landed in it. Since then the gate fails on `test__fast_mail_parser___parse_qp_message_metadata` at **+7.3%**, while the change this PR actually makes reads **−96.4%** on `headers_repeat_read`, exactly as intended.

Nothing here can reach quoted-printable decoding. The same thing happened to #247 on the same rebase — a pure thread-scheduler change, which came back **+19.0%** on `parse_qp_message` with per-round minima of 0.279 / 0.280 / 0.278 — and to #248, where I burned three `#[inline(never)]` attempts moving the number 0.268 → 0.265 → 0.277 against an unmoving base. Three PRs, three unrelated changes, one benchmark.

So this is code placement, not this diff, and #249 is the instrument for it: it builds the same source K times differing only in a layout salt and reports the per-benchmark spread across salts. If `parse_qp_message*` turns out to have a spread near or above 7%, then a 7.3% verdict on it is not evidence about this PR — and the right fix is upstream of all three, not a per-PR attribute.

I would rather leave this failing and explained than make it pass with a guess. Happy to rebase again once #249 has been dispatched and there is a measured number to judge it against.

PR #247, opened 2026-09-17
Size parse_many's workers by bytes, and let the caller be one (#232)

### LINKED ISSUE #232: parse_many: run the claim loop on the calling thread, gate workers on batch bytes, and stop probing available_parallelism() per call
## Problem

`parse_many` (all three modes) spawns a fresh set of OS threads on **every call** and sizes that set only by *message count*, never by how much work there is. A batch of 16 one-kilobyte messages -- an IMAP fetch page, a queue poll -- spawns `min(cores, 16)` threads, each of which parses roughly one message (~1 µs at the README's ~1.1 µs/KB) and exits. A `pthread_create` + join is on the order of 15-40 µs (stack mmap, clone, scheduler wake), so for that batch the scheduler costs one to two orders of magnitude more than the parsing it schedules, and `parse_many(batch)` is expected to be markedly **slower than `parse_many(batch, threads=1)`** for exactly the shape a mail pipeline produces most often. The README currently tells users the opposite: `parse_many` "is the right default for a mail pipeline ... and it costs nothing at any size" (README.md:344-346).

Two smaller costs ride along. The calling thread -- which already sits with the GIL released inside `py.detach` -- does no parsing; it only blocks on the joins, so every call pays one spawn/join more than it needs (for `workers == 2` that is half the thread cost). And `std::thread::available_parallelism()` is evaluated on every call whose `threads` is `None`; on Linux std's implementation reads `/proc/self/cgroup`, `/proc/self/mountinfo` and the cgroup `cpu.max` files each time (not memoised in std as of the pinned 1.98.0 -- confirm in the toolchain's `library/std/src/sys/pal/unix/thread.rs` before quoting a number), which is several file opens per `parse_many` call on the platform the CI gate and production run on.

For the batches the benchmark suite does measure (2000 × 0.8 KB, 16 × 767 KiB) the cost is amortised and expected to be a few percent at most. The shape where it dominates -- a small batch with the default thread count -- has no benchmark at all, which is why it has gone unnoticed.

## Evidence

- `src/mail_parser.rs:361-365` -- workers are capped on message count only, and the parallelism probe runs per call:
  ```rust
  let available = threads
      .or_else(|| thread::available_parallelism().ok())
      .map_or(1, NonZeroUsize::get);
  // Never more workers than there is work for them to do.
  let workers = available.min(payloads.len()).max(1);
  ```
  "Work" here means messages, not bytes: 16 × 1 KB gets as many workers as 16 × 1 MB.
- `src/mail_parser.rs:375-395` -- `workers` fresh scoped threads per call, none reused:
  ```rust
  let handles: Vec<_> = (0..workers)
      .map(|_| {
          scope.spawn(|| {
              let mut mine = Vec::new();
              loop {
                  let index = cursor.fetch_add(1, Ordering::Relaxed);
                  if index >= payloads.len() { break; }
                  mine.push((index, parse(payloads[index].as_ref())));
              }
              mine
          })
      })
      .collect();
  ```
- `src/mail_parser.rs:396-404` -- the calling thread only joins: `.map(|handle| match handle.join() { Ok(results) => results, Err(payload) => std::panic::resume_unwind(payload) })`. It never claims an index.
- `src/mail_parser.rs:317-323` -- the design is deliberately dependency-free: "Uses `std::thread::scope` and a shared atomic cursor rather than a thread pool crate. Two reasons: it adds no dependency ... and the cursor gives *dynamic* work distribution". Any fix must stay std-only and keep the dynamic cursor.
- `src/mail_parser.rs:8-19` -- the core forbids shared state: "no `static mut`, no `OnceCell`/`OnceLock`, no `thread_local` ... Introducing shared mutable state here invalidates that audit and must re-open it." So caching the default parallelism cannot live in the core.
- `src/fast_mail_parser.rs:1519-1525` -- the binding layer passes `None` straight through, so nothing caches the default:
  ```rust
  fn resolve_workers(threads: Option<usize>) -> PyResult<Option<NonZeroUsize>> {
      match threads {
          Some(0) => Err(exceptions::PyValueError::new_err(
              "threads must be at least 1; pass threads=None for the default",
          )),
          other => Ok(other.and_then(NonZeroUsize::new)),
      }
  }
  ```
  The binding layer already uses `OnceLock` (`src/fast_mail_parser.rs:25`, `:517`, `:963`), so a process-wide cache there has precedent and does not touch the core's audit.
- `src/fast_mail_parser.rs:1548` -- `let parsed = py.detach(|| mail_parser::parse_many(&messages, workers));` -- the caller is already off the GIL for the whole batch, so it is free to parse.
- `README.md:329-335` -- the documented cost model: "~24 µs of overhead per message against a parse that costs roughly 1.1 µs per KB". At that rate a thread spawn costs as much as parsing ~15-40 KB.
- `README.md:163` -- 2000 × 0.8 KB, all cores: **4.67 ms** on the 4-vCPU runner. Four spawns plus a cgroup probe is ~100-200 µs of that, i.e. the amortised case is only 2-4%.
- `README.md:344-346` -- "`parse_many` is the right default for a mail pipeline, where messages are usually a few KB, and it costs nothing at any size" -- not true for a small batch with `threads=None` today.
- `tests/benchmark/test_read_message.py:157-165` -- the only fixed-thread `parse_many` benchmark pins `threads=1` ("Single-threaded on purpose ... thread scheduling would only add noise"); `:184-189` (16 × 767 KiB) and `:227-232` (`SMALL_BATCH = 2000`) use the default. No benchmark covers a small batch with the default thread count.
- `tests/test_parse_many.py:8` and `fast_mail_parser/__init__.pyi:663` -- the contract: "`threads` caps the worker count; the default is the machine's parallelism." A *cap* -- using fewer workers when there is less work is within it; `tests/test_parse_many.py:217-222` already asserts a 1-message batch with `threads=32` "must not spawn idle workers".
- Known profile for context: a full parse of `tests/data/large_message.eml` is 0.225 ms (M4) / 0.604 ms (EPYC) and metadata mode 0.030 ms; a 1 KB message is ~1 µs. None of the per-message hot loops (base64 decode 53%, whitespace strip 27%, memmem 7.5%) are touched by this change.

## Proposal

Three std-only changes to the scheduler, no new dependency, results and error semantics identical.

**(a) The calling thread is a worker.** Spawn `workers - 1` scoped threads and run the identical claim loop on the calling thread inside the `thread::scope` closure; its `Vec` joins the spawned workers' results. One fewer spawn/join per call. A panic on the calling thread propagates out of `thread::scope` (which still joins the other workers first) and is caught by `catch_panics` at `src/fast_mail_parser.rs:175-187` exactly as a `resume_unwind` from a joined handle is today.

**(b) Size workers by bytes as well as by count.**

```rust
/// Below this many input bytes per worker, starting the worker costs more than
/// it saves: ~1.1 µs/KB of parse (README) against ~15-40 µs to create and join
/// an OS thread puts break-even near 30 KB; 64 KiB leaves margin for slower
/// encodings and slower spawns. A batch with less work than this runs inline.
const MIN_BYTES_PER_WORKER: usize = 64 * 1024;

let total_bytes: usize = payloads.iter().map(|p| p.as_ref().len()).sum();
let by_bytes = (total_bytes / MIN_BYTES_PER_WORKER).max(1);
let workers = available.min(payloads.len()).min(by_bytes).max(1);
```

With `threads=None` a 16 × 1 KB batch runs inline on the caller (the existing `workers == 1` fast path at `src/mail_parser.rs:367-372`); 2000 × 0.8 KB (1.6 MB) still gets `min(available, 25)`; 16 × 767 KiB is unaffected. An explicit `threads=N` remains an upper bound only -- the byte gate may lower it, never raise it.

**(c) Cache the default parallelism in the binding layer.** In `src/fast_mail_parser.rs`, `resolve_workers` fills `None` from `static DEFAULT_WORKERS: OnceLock<NonZeroUsize>` initialised once from `thread::available_parallelism().unwrap_or(NonZeroUsize::MIN)`. The core's `.or_else(available_parallelism)` fallback at `:362` stays for the fuzz/by-path consumers but is no longer reached from Python. The core's no-shared-state invariant is untouched.

Behaviour that must stay identical: one result per input in input order; `Ok`/`Err` per slot; `raise_on_error`, `strict`, `mode=`; `threads=0` -> `ValueError`; negative -> `OverflowError`; the GIL released for the whole batch; a parser panic surfaces as `ParseError("internal parser panic: ...")`.

1. `src/mail_parser.rs`, `parse_many_as` (`:348-415`): add `MIN_BYTES_PER_WORKER` next to the DoS constants (`:47-58`) with the reasoning comment; compute `total_bytes`/`by_bytes`; extract the claim loop into a small generic `fn` (or a `Fn` closure capturing `&cursor`, `payloads`, `parse` -- all `Sync`) so the spawned workers and the caller run the same code; spawn `1..workers`; run the loop on the caller; collect `own` plus the joined handles. Update the doc comments at `:317-328` ("Callers with a batch smaller than the thread count do not spawn idle workers" -> also "a batch with less than `MIN_BYTES_PER_WORKER` of work per worker uses fewer, and below one worker's worth runs on the calling thread") and the panic comment at `:396-401`. Keep the file PyO3-free -- it is compiled by path into `fuzz/fuzz_targets/parse_email.rs:22` and `parse_agreement.rs:48`.
2. `src/fast_mail_parser.rs`, `resolve_workers` (`:1519-1525`): add the `OnceLock` and fill `None`; extend the doc comment with why the cache lives here and not in the core (cite the module doc at `src/mail_parser.rs:8-19`). Note the accepted trade-off: a cgroup CPU quota changed after the first call is not observed.
3. `fast_mail_parser/__init__.pyi:663-665` (and the three mode overloads at `:700`, `:711`, `:722` if they repeat the sentence): "`threads` caps the worker count; the default is *at most* the machine's parallelism, and small batches use fewer workers -- below ~64 KiB of input the batch parses on the calling thread." Do not change the signature; `tests/test_contract.py:34` freezes the name only. While there, the sentence at `:670-672` ("its one cost -- copying each payload before parsing begins") describes the pre-#96 copy and is stale; fix it in the same pass or leave it and say so in the PR.
4. `README.md:344-346`: qualify "costs nothing at any size" -- after this change it is true for small batches too, so the wording can stay once measured, but say that small batches run inline. Also consider a line under `README.md:312-318` ("When it helps").
5. Tests (oracle): `tests/test_parse_many.py` -- everything under `# --- threads` (`:205-222`), `test__gil_is_released_for_the_duration_of_the_batch` (`:227-256`; its 400 × ~2.2 KB batch is ~880 KB, so it still gets >1 worker -- keep it that way), `test__absurdly_large_thread_cap_is_harmless` (`:272-275`), `test__threads_still_caps_the_workers_in_every_mode` (`:390`), and the full-corpus agreement tests (`:103-140`, `:358-380`). Add:
   - `test__a_small_batch_with_default_threads_matches_threads_one`: 16 × `_message(...)` (well under 64 KiB), `parse_many(p)` == `parse_many(p, threads=1)` subject-for-subject, in every mode.
   - `test__a_large_batch_with_default_threads_still_matches_threads_one`: 400 × (`_message(...) * 20`) -- above the byte gate on any core count -- same equality, so the parallel path stays covered.
   - `test__threads_is_an_upper_bound_not_a_target`: `parse_many(small_batch, threads=64)` returns correct results (it may run inline; the assertion is on results only, since worker count is not observable from Python).
6. Benchmarks in `tests/benchmark/test_read_message.py`, after `SMALL_BATCH`/`_small_message()` (`:215-232`): add `SMALL_PAGE = 16` and two informational `test__threaded___` benchmarks: `test__threaded___parse_many_small_page` (`parse_many([_small_message()] * SMALL_PAGE)`) and `test__threaded___parse_many_small_page_threads1` (same batch, `threads=1`). Both prefixed `test__threaded___` so the gate reports and does not judge them (`:168-181`). `check_benchmark.py` selects by exact name (`.github/scripts/check_benchmark.py:62-63`), so new names cannot collide with the gate.
7. `CHANGELOG.md` under `## [Unreleased]` -> `### Changed`: one entry with the local before/after numbers and the runner's.
8. Traps: no change under `vendor/` is needed, so `PATCH.md` and the vendored tests are untouched; if you do touch `vendor/`, never run cargo there without `--target-dir` elsewhere (the sdist include glob sweeps `target/` in). The crate is abi3-py311; nothing here touches the Python ABI. `clippy` runs with `-W clippy::cast_possible_truncation`; `total_bytes / MIN_BYTES_PER_WORKER` is `usize`/`usize`, so no cast is needed -- do not introduce one.
9. Out of this change: guided chunking of the cursor (`fetch_add(k)`) -- plausible 1-2% on the 2000-message batch, unmeasured; a separate issue if the A/B below shows cursor contention.

## Harness -- how to prove it

Build both sides once, then measure interleaved (the `ab_median.py` pattern -- separate build-and-measure cycles read ~16% apart on a no-op, `.github/scripts/ab_median.py:8-19`):

```sh
# A = master 73f56da, B = the branch; two venvs, one wheel each
uv venv .venv-a && uv venv .venv-b
for s in a b; do uv pip install --python .venv-$s/bin/python maturin pytest pytest-benchmark==5.3.0 mail-parser==4.6.4; done
git worktree add ../fmp-a 73f56da
(cd ../fmp-a && ../fmp/.venv-a/bin/maturin develop --release)
.venv-b/bin/maturin develop --release
for r in 1 2 3; do
  for s in a b; do
    .venv-$s/bin/pytest -q tests/benchmark --benchmark-min-rounds=25 --benchmark-json=$s-$r.json
  done
done
AB_THRESHOLD_PCT=7 python .github/scripts/ab_median.py --a-label master --b-label branch \
  --a a-1.json a-2.json a-3.json --b b-1.json b-2.json b-3.json
```

Note the new benchmarks only exist on side B; either run side B's test file against both venvs (what `test.yml:427-449` does) or read them from the B reports only.

Expected, to be confirmed by the numbers:
- `test__threaded___parse_many_small_page` (new, 16 × ~0.8 KB, default threads): today several times slower than `..._threads1`; after, within ~1.2x of it. This is the headline number for the PR.
- `test__threaded___parse_many_small` (2000 × 0.8 KB): -1% to -4% (one fewer spawn, no per-call cgroup probe). Below the gate's resolution; report it, do not claim more than measured.
- `test__threaded___parse_many` (16 × 767 KiB): 0 to -1%.
- `test__fast_mail_parser___parse_many` and `..._parse_many_metadata` (`threads=1`): unchanged (the `workers == 1` fast path is the same code).
- `test__fast_mail_parser___parse_message` (the gate) and every decoding benchmark: 0. No benchmark may regress more than 7%; the pure-Python controls (`test__mail_parser___*`, `test__stdlib_email___*`) define the noise floor and any treatment delta must clear it.

The M4 is not a proxy for x86: code-layout shifts on Zen have moved the decode benchmarks up to 96% for byte-identical source (`ab_median.py:56-61`), and this change touches the hot crate. The PR's CI benchmark job (interleaved A/B, `test.yml:323-475`) is the x86 verdict; if it shows an unexplained move on `parse_message`, run `toolchain-ab.yml` against the branch before concluding anything.

Full local surface, all of which must pass:

```sh
cargo fmt --all -- --check
ruff check .
```

Put the local before/after table (both new benchmarks, the three existing `parse_many` benchmarks, the gate benchmark, and the control delta) in the PR body, with the CPU it was measured on.

## Acceptance criteria

- [ ] `parse_many_as` runs the claim loop on the calling thread and spawns `workers - 1` threads; the atomic cursor and dynamic distribution are unchanged; no new dependency in `Cargo.toml`/`Cargo.lock`.
- [ ] `MIN_BYTES_PER_WORKER` is a named `const` in `src/mail_parser.rs` with a comment giving the cost model it is derived from; worker count is `min(available, len, total_bytes / MIN_BYTES_PER_WORKER).max(1)`; an explicit `threads=` is never exceeded.
- [ ] The default parallelism is resolved once per process via `OnceLock` in `src/fast_mail_parser.rs`; `src/mail_parser.rs` still contains no `OnceLock`/`OnceCell`/`static mut`/`thread_local`/`unsafe` (grep-checkable) and still compiles by path in `fuzz/`.
- [ ] `tests/test_parse_many.py` gains the three tests named above and the whole file passes, including `test__gil_is_released_for_the_duration_of_the_batch`.
- [ ] `tests/benchmark/test_read_message.py` gains `test__threaded___parse_many_small_page` and `test__threaded___parse_many_small_page_threads1`; on the branch the default-threads variant is within 1.2x of the `threads=1` variant (median of 3 interleaved rounds, reported in the PR body).
- [ ] Interleaved A/B against `73f56da`: no benchmark slower by more than 7%; `test__threaded___parse_many_small` not slower; `test__fast_mail_parser___parse_message` within the control noise floor on the CI x86 runner.
- [ ] `__init__.pyi` `threads` docstring(s) and `README.md:344-346` describe the byte gate; `CHANGELOG.md` `[Unreleased]` has the entry with measured numbers.
- [ ] `pytest tests --ignore=tests/benchmark`, `cargo fmt --check`, `cargo clippy` (flags above), `mypy --strict fast_mail_parser/`, `ruff check .` all green.

## Out of scope

- A persistent thread pool or any pool crate -- the no-dependency design at `src/mail_parser.rs:317-323` stands.
- Guided/chunked cursor claims (`fetch_add(k)`) -- separate issue if the A/B shows contention on the 2000-message batch.
- Changing what `threads=None` means beyond "at most the machine's parallelism"; changing `threads=0`/negative handling.
- Any change under `vendor/mailparse` or to the per-message decode loops.
- Re-measuring the README's per-KB cost model; the constant is derived from the published number and tuned by the harness above.

## Related issues

- #224 — the thread-scaling benchmarks that prove this
- #223

kurok:
Closes #232.

Branched from `master` (`e7bbb05`). Independent of #245 and #246.

## The bug, measured before the fix

`parse_many` capped its workers on **message count alone**, never on how much work there was. Sixteen one-kilobyte messages — an IMAP fetch page, a queue poll — got `min(cores, 16)` threads, each of which parsed roughly one message and exited. Creating and joining an OS thread is 15–40 µs against a parse of ~1.1 µs/KB, so the scheduler cost one to two orders of magnitude more than the work it scheduled.

On this machine, **on master**:

| | min | mean |
|---|---|---|
| `parse_many(page)` — default threads | 63.5 µs | 108.4 µs |
| `parse_many(page, threads=1)` | 28.4 µs | 31.2 µs |

**`parse_many` was 2.2x slower than asking it not to use threads** — for the shape a mail pipeline produces most often, against a README that says it "costs nothing at any size". No benchmark covered that shape, which is why it went unnoticed.

## What changed

Three std-only changes. No new dependency, the dynamic cursor kept.

**(a) Workers are capped by bytes as well as by count.** `MIN_BYTES_PER_WORKER = 64 KiB`; a batch with less total work than one worker's worth takes the existing inline path. `threads=` stays an upper bound — the gate may lower it, never raise it.

**(b) The calling thread is a worker.** The claim loop is now a closure that both the spawned workers and the caller run; `workers - 1` get spawned. The caller already sits inside `py.detach` with the GIL released, so blocking it on joins was one spawn and one join per call for nothing — half the thread cost when `workers == 2`. A panic on the caller unwinds out of `thread::scope`, which joins the spawned workers first, and reaches `catch_panics` exactly as a `resume_unwind` from a joined handle did.

**(c) The default parallelism is probed once per process**, in the binding layer. `std` does not memoise `available_parallelism`; on Linux it reads `/proc/self/cgroup`, `/proc/self/mountinfo` and the cgroup `cpu.max` files on *every* call. The cache cannot live in `mail_parser` — that module's docs rule out statics and `OnceLock` so its free-threading audit stays valid — so it lives next to the `OnceLock`s this layer already has, and the core keeps its own fallback for the fuzz targets that include it by path. The trade-off is stated at the function: a cgroup quota changed after the first call is not observed; pass `threads=` to override.

## Numbers

Apple M4 (10 vCPU), rustc 1.98.0, CPython 3.12, 5 interleaved rounds. Pure-Python controls within **0.5%**.

|---|---|---|---|
| `parse_many_small_page` *(new)* | 0.070 ms | **0.032 ms** | **2.2x** |
| `parse_many_small_page_threads1` *(new control)* | 0.029 ms | 0.028 ms | +1.0% flat |
| `parse_many` (16 × 767 KiB) | 1.326 ms | 1.333 ms | −0.5% flat |

**The control is the point.** Same batch, same parsing, scheduler forced off — it does not move. That is what says the 2.2x is scheduling and not parsing, and it is why the benchmark was added in a pair.

The 16 × 767 KiB batch is unaffected, as intended: 12 MB is far above the gate.

**What I am *not* claiming from this table.** `parse_many_small` and `parse_many_metadata_small` read ~27% faster, `threaded___parse_many` ~8%, and every metadata/lazy-untouched benchmark +16–24%. Those last ones are the M4 code-placement artifact documented on #244 — master sits at 0.037 ms on `parse_metadata` where it was 0.030 before base64-simd was linked, and x86 measured them flat on #244, #245 and #246. The 2000-message rows are above the byte gate and spawn the same worker count either way, so I do not have a mechanism that accounts for 27%; the plausible one (one fewer spawn out of ten) is worth well under 1%. **Treat those rows as unexplained until the gate reports.** The claim this PR makes is the first row and its control.

| Benchmark | workers | master | Delta | |
|---|---|---|---|---|
| `test__fast_mail_parser___full_read` | 0.163 ms | 0.171 ms | +4.7% |  |
| `test__fast_mail_parser___parse_lazy_all_attachments` | 0.174 ms | 0.181 ms | +3.8% |  |
| `test__fast_mail_parser___parse_lazy_untouched` | 0.043 ms | 0.050 ms | +16.5% |  |
| `test__fast_mail_parser___parse_many` | 1.333 ms | 1.326 ms | -0.5% |  |
| `test__fast_mail_parser___parse_many_metadata` | 0.241 ms | 0.294 ms | +21.9% |  |
| `test__fast_mail_parser___parse_message` | 0.157 ms | 0.163 ms | +4.0% |  |
| `test__fast_mail_parser___parse_message_strict` | 0.157 ms | 0.163 ms | +4.0% |  |
| `test__fast_mail_parser___parse_metadata` | 0.030 ms | 0.037 ms | +24.0% |  |
| `test__fast_mail_parser___parse_metadata_str` | 0.030 ms | 0.037 ms | +23.5% |  |
| `test__fast_mail_parser___parse_tree` | 0.157 ms | 0.165 ms | +5.0% |  |
| `test__fast_mail_parser___parse_tree_lazy_untouched` | 0.041 ms | 0.049 ms | +17.0% |  |
| `test__fast_mail_parser___parse_tree_metadata` | 0.031 ms | 0.038 ms | +22.9% |  |
| `test__mail_parser___parse_message` | 5.476 ms | 5.497 ms | +0.4% | control |
| `test__mailparser_lib___full_read` | 5.790 ms | 5.821 ms | +0.5% | control |
| `test__stdlib_email___full_read` | 7.450 ms | 7.429 ms | -0.3% | control |
| `test__threaded___parse_many` | 0.677 ms | 0.731 ms | +8.0% | informational |
| `test__threaded___parse_many_metadata_small` | 1.975 ms | 2.521 ms | +27.7% | informational |
| `test__threaded___parse_many_small` | 2.187 ms | 2.789 ms | +27.5% | informational |
| `test__threaded___parse_many_small_page` | 0.032 ms | 0.070 ms | +123.2% | informational |
| `test__threaded___parse_many_small_page_threads1` | 0.028 ms | 0.029 ms | +1.0% | informational |
| `test__threaded___threadpool_parse_email` | 0.631 ms | 0.647 ms | +2.5% | informational |
| `test__threaded___threadpool_parse_email_small` | 14.095 ms | 14.203 ms | +0.8% | informational |

Noise floor from the pure-Python controls: **0.5%** (they cannot be affected by the build, so this is measurement error).

</details>

## Contract

Unchanged: one result per input in input order, per-slot `Ok`/`Err`, `raise_on_error`, `strict`, `mode=`, `threads=0` → `ValueError`, negative → `OverflowError`, the GIL released for the whole batch, a parser panic surfacing as `ParseError`.

Worker count is not observable from Python, so the new tests in `tests/test_parse_many.py` pin the only thing that is — that results do not depend on how the scheduler sized itself:

- `test__a_small_batch_with_default_threads_matches_threads_one` — 16 messages, asserted under the gate, all three modes.
- `test__a_large_batch_with_default_threads_still_matches_threads_one` — 400 × 20 messages, asserted *above* the gate, so the genuinely parallel path stays covered rather than being quietly serialised by the change under test.
- `test__threads_is_an_upper_bound_not_a_target` — `threads=64` on a tiny batch still returns correct results.
- `test__a_single_message_batch_still_parses` and `test__empty_and_tiny_batches_do_not_divide_by_zero` — `workers` bottoms out through two `.min()`s and a `.max(1)`, and `total_bytes / MIN_BYTES_PER_WORKER` is integer division, so these are where an off-by-one or a zero would show.

## Docs

`Readme.md` said `parse_many` "costs nothing at any size". That was false for small batches, so the section now says workers are sized by bytes, that a sub-64-KiB batch parses inline, and what it used to cost. The `parse_many` docstring in `__init__.pyi` says `threads` is an upper bound and that small batches use fewer.

## Checks run locally

`pytest tests --ignore=tests/benchmark` (820 passed, 3 skipped), `cargo fmt --all --check`, `cargo clippy --all-targets -- -D warnings -W clippy::cast_possible_truncation`, `mypy --strict`, `ruff check .` — all green. No `vendor/` change, so no `PATCH.md` and no vendored `cargo test`. `src/mail_parser.rs` stays PyO3-free (it is compiled by path into two fuzz targets) and gains no shared state — the one cache added is in the binding layer.


kurok:
**Gate status, and why I am not patching it here.**

This PR was green at 15/15 before I rebased it onto `2aee3df`. The only change since is the base — #229's quoted-printable decoder landed in it — plus a one-line `CHANGELOG.md` conflict resolution. The gate now fails on `test__fast_mail_parser___parse_qp_message` at **+19.0%**.

**This PR is the clearest evidence in the series that the failure is not the PR.** It touches `src/fast_mail_parser.rs` and `src/mail_parser.rs` only — the thread scheduler. There is no path from a worker-sizing change to a quoted-printable decoder it never calls. And the run says as much itself:

- `parse_qp_message_metadata` — same message, same headers, decode skipped — **−0.3%, flat**
- `parse_many_small_page`, what this PR is for — **−55.6%**, with its forced-serial control at +1.4%
- `parse_qp_message` — **+19.0%**, per-round minima 0.279 / 0.280 / 0.278

Rock-stable across rounds, so not noise; and the only work separating the flat row from the failing one is the decode.

#246 failed the same way on the same rebase (+7.3%), and #248 took three `#[inline(never)]` attempts that moved its number 0.268 → 0.265 → 0.277 against an unmoving 0.246 base before I stopped guessing. #249 adds the instrument: same source, K builds differing only in a layout salt, per-benchmark spread across salts. That number is what a 19% verdict here should be read against.

Leaving it failing and explained rather than passing by guesswork. Ready to rebase once there is a measured spread to judge against.

kurok:
**Rebased onto `c403a69`. The code did not change. The gate's verdict changed completely — and that is the answer.**

Same source as the previous run (a `CHANGELOG` line and a hand-merged benchmark file aside), different base commit, therefore a different binary layout. Compare the two runs on x86:

| Benchmark | previous run | this run |
|---|---|---|
| `parse_qp_message` | **+19.0%** (failed the gate) | **−6.9%** |
| `parse_qp_message_metadata` | −0.3% | −10.4% |
| `parse_tree_lazy_untouched` | — | **+15.2%** (fails the gate) |
| `parse_many_metadata` | — | **+11.9%** |
| `parse_lazy_untouched` | — | **+11.1%** |
| `parse_tree` | — | **+9.3%** |
| `parse_metadata_str` | — | **−9.7%** |
| controls | 3.5% | **0.5%** |

The benchmark that fails **moved**. The one that failed at +19% is now 6.9% *faster*. The same unchanged scheduler change now scatters ±15% in both directions against a 0.5% control floor.

And #246 — rebased in the same batch, also no code change — **went from failing at +7.3% to passing 15/15**.

Two independent flips in one batch. This is not a property of either PR; it is the spread of the instrument. On this revision, on these runners, code placement alone moves the parse benchmarks by at least ±15%, which is more than twice the gate's 7% threshold. A 7% verdict on these paths is currently a coin toss with extra steps.

That is exactly the number #249 exists to measure properly — four salted builds of one source, interleaved, reporting the per-benchmark spread — instead of us inferring it from failures. The inference is now strong enough to act on: **this PR should be judged against the measured spread, not against 7%**, and I am not going to touch its code to make a placement lottery land the right way.

Its own benchmark, meanwhile, is doing exactly what it was written to do: `parse_many_small_page` **−44.2%**, with the forced-serial control at −0.4%.

Issue #240, opened 2026-08-30
Measure x86 code-layout sensitivity with a salted layout A/B, and trial 64-byte block alignment for x86_64 wheels


kurok:
## Problem

The PR benchmark gate compares two builds of the extension on one x86 runner (`.github/workflows/test.yml` benchmark job, `AB_THRESHOLD_PCT: "7"`). Two builds that differ at all differ in code layout, and on the Zen CPUs GitHub's fleet provides, layout alone has been measured to move the decode-path benchmarks by +3.2-4.4% on the gate's own CPU (#205), +3-7% on an Apple M4 at the 0.9.0 cut (CHANGELOG.md:27-30), and +27% / +96% on one runner draw for a binary whose x86-64 instruction stream was byte-identical (#207, CONTRIBUTING.md:328-336). The two byte-at-a-time loops that produced the 96% class are fixed (#213, #214), but what remains of a full parse is 53% `data_encoding::decode_base_mut`, 27% the inlined whitespace strip and 7.5% `memmem` (M4 profile) -- loops whose addresses the fat-LTO link decides -- and CONTRIBUTING.md:359-360 records that "nobody has measured how sensitive they are".

Today the gate's only remedy for a suspicious verdict is prose: `ab_median.py:193-200` tells the reader to **re-run** and compare the CPU line, i.e. hope the next runner draw is a different CPU model. A maintainer whose local machine is an M4 cannot reproduce the x86 effect at all, and QEMU is not a proxy (no micro-op-cache model). Every 7%+ verdict on a decode-path change -- and three round-1 issues (*Decode base64 with base64-simd*, *run-copying QP decoder*, *Evaluate Body once per part*) are exactly that -- is therefore "maybe".

The hardware is not the missing piece; the CI runners already are x86 Zen. What is missing is a **controlled experiment**: the same source built K times differing only in a layout salt, measured interleaved on one runner. The per-benchmark spread across salts *is* that revision's layout sensitivity on that CPU. The same harness, run once more with x86 code-alignment rustflags, decides whether the tail can be removed from the shipped x86_64 wheels (median unchanged, 0-96% tail gone) or whether that idea should be recorded as tried and rejected.

## Evidence

All citations are from master `73f56da`.

- `CONTRIBUTING.md:305-309` -- "**This crate is unusually sensitive to codegen.** A rustc minor version alone moved the parse path 15-96% (#120) ... the lesson stands: do not assume a change is free because it looks free."
[...]

PR #248, opened 2026-09-17
Stop allocating header keys twice, and test for a header without decoding it (#238)

### LINKED ISSUE #238: Header values: single-line fast path in the vendored `get_value`, drop the double key allocation in `collect_headers`
## Problem

Every header value this library reads goes through mailparse's two-stage tokenizer: `tokenize_header` splits the value into lines and allocates a `Vec<HeaderToken>` per line plus an outer `Vec`, `normalize_header_whitespace` builds a second `Vec`, and `normalize_header` then grows a `String` from the tokens with `push_str` and no capacity hint. That machinery exists for folded values and RFC 2047 encoded words. The common case is neither: on `tests/data/large_message.eml` the root block has 24 headers, 8 folded continuation lines and exactly one encoded word (the Subject), and every part-level `Content-Type`, `Content-Transfer-Encoding`, `Content-Disposition`, `Content-ID`, `Date`, `MIME-Version`, `X-*` value is one physical line with no `=?`.

The tokenizer is paid more than once per header. `collect_headers` calls `get_value()` for all ~35 headers of the fixture; mailparse itself then calls `get_first_value` -- a linear scan that re-tokenises the matching header -- for `Content-Type` on every part, `Content-Transfer-Encoding` on every `get_body_encoded()` (two to three times per leaf), `Content-Disposition` on `get_content_disposition()`; and our three flat parsers add `Subject`, `Date`, `Content-ID` and a second `Content-Disposition` lookup that exists only to test presence. That is roughly 60 `get_value` calls per full parse, ~55 of them for values the fast path below handles with a single `to_owned()`. `collect_headers` additionally allocates the key twice (`get_key()` then `key.clone()` for the position map) when the raw key bytes would do.

Why it matters for a library whose reason to exist is speed: in `mode="metadata"` (0.030 ms on the M4 fixture) the boundary `memmem` accounts for roughly 17-20 us, so the header path is a large share of what is left; and on the 2000 x ~0.8 KiB batch the benchmarks call the mailbox-sweep case (#96), header work is essentially all the Rust-side work per message. None of the known profile hot spots (`decode_base_mut` 53%, whitespace strip 27%, `memmem` 7.5%) touches this; it is the next layer down, and it is the layer every mode pays.

## Evidence

- `vendor/mailparse/src/lib.rs:213-240` -- `pub fn get_value(&self) -> String { let chars = self.decode_utf8_or_latin1(); self.normalize_header(chars) }` and `fn normalize_header(...) -> String { let mut result = String::new(); for tok in header::normalized_tokens(&chars) { ... result.push_str(t) ... } }`. Every value is tokenised; the result `String` starts empty. `get_value_utf8` (`lib.rs:258-263`) goes through the same `normalize_header`.
- `vendor/mailparse/src/header.rs:137-151` -- `fn tokenize_header(value: &str) -> Vec<HeaderToken> { let mut tokens = Vec::new(); let mut lines = value.lines(); ... while let Some(line) = lines.next().map(str::trim_start) { ... let mut line_tokens = tokenize_header_line(line); tokens.append(&mut line_tokens); } }`: one `Vec` per physical line plus the outer one.
- `vendor/mailparse/src/header.rs:75-129` -- `tokenize_header_line`: `match find_from(line, ix_search, "=?") { ... None => { result.push(maybe_whitespace(&line[ix_search..])); break; } }` (122-125). With no `=?` in the line the whole (already `trim_start`ed) line is one `Text` or `Whitespace` token. `find_from` is `line[ix_start..].find(key)` (`lib.rs:118-120`).
- `vendor/mailparse/src/header.rs:160-217` -- `normalize_header_whitespace`: `let mut result = Vec::<HeaderToken>::new();` (161); a lone `Text` is pushed unchanged (168-179), a lone `Whitespace` is pushed unchanged because `saved_token` is `None` (180-191); `" ".to_string()` at 196/201/203 allocates per fold. `normalized_tokens` at 219-221 chains the two.
- `vendor/mailparse/src/lib.rs:371-377` -- `HeaderParseState::Value => { if c == b'\n' { state = ValueNewline } else if c != b'\r' { ix_value_end = ix + 1; } }`: a header value slice never ends in `\r` (only a non-CR non-LF byte advances `ix_value_end`), so `str::lines()` on a value with no `\n` yields exactly the value itself, once (or nothing for `""`).
- `vendor/mailparse/src/lib.rs:464-472` -- `fn get_first_value(&self, key: &str) -> Option<String> { for x in self { if x.get_key_ref().eq_ignore_ascii_case(key) { return Some(x.get_value()); } } None }`: each lookup re-runs the tokenizer on the hit.
- Callers inside mailparse, per part: `lib.rs:943-947` `headers.get_first_value("Content-Type")`; `lib.rs:825-832` `get_body_encoded`: `self.headers.get_first_value("Content-Transfer-Encoding").map(|s| s.to_lowercase())`; `lib.rs:845-850` `get_content_disposition`: `self.headers.get_first_value("Content-Disposition")`.
- Callers in this crate: `src/mail_parser.rs:1312-1319` (`get_first_value("Subject")`, `("Date")`), `1356` (`part.get_content_disposition()`), `1386-1389` (`get_first_value("Content-ID")`), `1396` (`disposition_token`); the same shape in metadata mode at `603-611`, `640`, `656-658` and lazy mode at `773-781`, `807`, `827-834`; tree at `948-954`, `1121-1127`.
- `src/mail_parser.rs:473-481` -- `fn disposition_token(part, kind) -> Option<String> { part.get_headers().get_first_value("Content-Disposition")?; Some(match kind { ... }) }`: line 474 builds and drops a normalised `String` purely to test presence; `get_first_header(..).is_some()` (`lib.rs:474-477`) answers the same question with no `get_value`.
- `src/mail_parser.rs:501-517` -- `collect_headers`: `let mut headers = Vec::new(); let mut positions: HashMap<String, usize> = HashMap::new(); for header in part.get_headers() { let key = header.get_key(); match positions.get(&key).copied() { Some(position) => headers[position].1.push(header.get_value()), None => { positions.insert(key.clone(), headers.len()); headers.push((key, vec![header.get_value()])); } } }`. `get_key()` is `decode_latin1(self.key).into_owned()` (`lib.rs:174-176`, whose doc says "Prefer using get_key_ref where possible for better performance"); `get_key_raw()` returns `&[u8]` (`lib.rs:273-275`). `part.get_headers()` iterates `&MailHeader` (`vendor/mailparse/src/headers.rs:51-58`); `ParsedMail.headers` is `pub` (`lib.rs:722`), so the count is known up front.
- `vendor/mailparse/src/addrparse.rs:286-291` -- `addrparse_header` calls `crate::header::normalized_tokens(&chars)` directly: the five address headers need the token stream and are unaffected by (and not slowed by) a fast path in `normalize_header`.
- `tests/data/large_message.eml` lines 1-32: root header block, separator at line 33; 8 continuation lines (4-5, 8-9, 11-13, 31); one `=?` (line 15). Part header blocks at 35-36, 39-40, 56-57, 145-149, 3946-3950. About 35 headers total, ~60 `get_value` calls per full parse counting the per-part lookups above.
- `vendor/mailparse/Cargo.toml:40-41` -- `[dependencies.memchr] version = "2.7.0"` is already a dependency of the vendored copy (`Cargo.lock:74-77` resolves 2.8.3); no new dependency is needed for the fast-path check.
- `vendor/mailparse/PATCH.md` -- "with **two functions changed** (one via a new module)" and the sync procedure ("keep `src/bytescan.rs`, the `mod bytescan;` line, the two call sites and `memchr`"): a change here makes it three and the file must say so.
- Known profile (do not re-measure): full parse 0.225 ms M4 / 0.604 ms EPYC; metadata 0.030 ms; remaining dominant loops `decode_base_mut` 53%, whitespace strip 27%, `memmem` 7.5%; everything else <5% each. `memmem` at 7.5% of 225 us is ~17 us, which in metadata mode leaves ~10-13 us for headers, per-part work and the Python call.

## Proposal

Three changes, no new dependency, no output change.

**1. Fast path in `normalize_header` (vendored, `vendor/mailparse/src/lib.rs`).** Before tokenising, take the single-line, no-encoded-word case directly:

```rust
fn normalize_header(&'a self, chars: Cow<'a, str>) -> String {
    let bytes = chars.as_bytes();
    // One physical line with no encoded word: `lines()` yields exactly this
    // string (a header value never ends in CR, see parse_header), the line
    // becomes one Text/Whitespace token, and normalize_header_whitespace
    // passes a lone token through unchanged. Same result as the tokenizer,
    // one allocation instead of ~6.
    if memchr::memchr2(b'\n', b'\r', bytes).is_none()
        && memchr::memmem::find(bytes, b"=?").is_none()
    {
        return chars.trim_start().to_owned();
    }
    self.normalize_header_tokens(chars)
}

fn normalize_header_tokens(&'a self, chars: Cow<'a, str>) -> String {
    let mut result = String::with_capacity(chars.len());
    for tok in header::normalized_tokens(&chars) { /* unchanged body */ }
    result
}
```

Equivalence argument, to go in the comment: `tokenize_header` calls `value.lines().map(str::trim_start)`; with no `\n` that is one line (none for `""`), and with the value never ending in `\r` the line is the whole value; `tokenize_header_line` with no `=?` pushes exactly one `maybe_whitespace(line)` token; `normalize_header_whitespace` pushes a lone `Text` or `Whitespace` unchanged. The fast path uses the same `str::trim_start` (Unicode `White_Space`) as `tokenize_header`. Checking `\r` as well as `\n` is redundant for values produced by `parse_header` but keeps the fast path correct for a `MailHeader` built any other way. `with_capacity(chars.len())` on the slow path is a hint, not a bound (decoded words can shrink or grow the value) -- it is fine either way. Keep `normalize_header_tokens` as a named function: it is the oracle for the differential test.

**2. `collect_headers` (`src/mail_parser.rs:501-517`).** Key the position map by raw bytes, size both containers, allocate the owned key once:

```rust
fn collect_headers(part: &ParsedMail<'_>) -> Vec<(String, Vec<String>)> {
    let count = part.headers.len();
    let mut headers: Vec<(String, Vec<String>)> = Vec::with_capacity(count);
    // Latin-1 decoding is injective, so raw-byte equality is exactly the
    // String equality this map had; keys stay case-sensitive as before.
    let mut positions: HashMap<&[u8], usize> = HashMap::with_capacity(count);
    for header in part.get_headers() {
        match positions.get(header.get_key_raw()).copied() {
            Some(position) => headers[position].1.push(header.get_value()),
            None => {
                positions.insert(header.get_key_raw(), headers.len());
                headers.push((header.get_key(), vec![header.get_value()]));
            }
        }
    }
    headers
}
```

Behaviour that must stay identical: first-appearance key order, all values kept per key, case-sensitive key grouping (`Received` and `received` remain separate entries, as today), Latin-1 fallback for non-UTF-8 key bytes via `get_key()`.

**3. `disposition_token` (`src/mail_parser.rs:473-481`).** Replace the presence test `part.get_headers().get_first_value("Content-Disposition")?;` with `part.get_headers().get_first_header("Content-Disposition")?;` -- same case-insensitive first-match semantics (`lib.rs:474-477`), no tokenizer, no `String`.

What this deliberately does not do: dedupe `get_body_encoded()` per part (round-1 issue "Evaluate Body once per part"), read Subject/Date out of the `collect_headers` table (the comment at `src/mail_parser.rs:1308-1311` explains #28; round-1 "Share one envelope/part-classification helper pair" is where that code moves), or rewrite `parse_header` (see Out of scope).

1. `vendor/mailparse/src/lib.rs`: split `normalize_header` into the fast-path dispatcher and `normalize_header_tokens` as sketched. `memchr` is already imported for `find_from_u8`; reuse the crate path. Do not touch `header.rs` -- the equivalence argument depends on its current behaviour, and every changed vendored file is one more hand-merge.
2. Add a differential test in `lib.rs`'s `mod tests` (it has `use super::*` and can construct `MailHeader { key, value }` directly): generate a corpus in the style of `vendor/mailparse/src/bytescan.rs:105-122` (xorshift over a small alphabet, lengths 0..80, several draws per length) with alphabet `b"ab =?\t\r\n_Q?B\xe9\xff"` so encoded-word delimiters, folds, tabs, Unicode-whitespace-after-Latin-1 and invalid UTF-8 all occur at every alignment; assert `h.get_value() == h.normalize_header_tokens(h.decode_utf8_or_latin1())` and the same for `get_value_utf8` where the bytes are valid UTF-8. Also assert a hand-written table: `""`, `"   "`, `"\tfoo"`, `"foo  "`, `"a\rb"`, `"=?"`, `"x =?utf-8?Q?y?= z"`, `"\u{a0}foo"` (NBSP is `White_Space`: both paths trim it).
3. Run the vendored suite the CI way: `cargo test --manifest-path vendor/mailparse/Cargo.toml --target-dir /tmp/vendor-target`. Never `cargo` inside `vendor/mailparse` without `--target-dir` elsewhere -- a `target/` there is swept into the sdist by the `vendor/mailparse/**/*` include (`pyproject.toml:46-48`).
4. `vendor/mailparse/PATCH.md`: change "two functions changed" to three, add a numbered item 3 describing `normalize_header` (what it skips, why the output is identical, which test proves it), and add it to the "Keeping this in sync" list. Also fix the stale "one function changed" in the root `Cargo.toml:53-55` comment while there (relates to round-1 "Make the vendored mailparse delta a CI invariant"; if that lands first, regenerate its `upstream.patch`).
5. `src/mail_parser.rs`: rewrite `collect_headers` and `disposition_token` as above. `HashMap<&[u8], usize>` borrows from `part`, which outlives the function -- no lifetime plumbing needed. Clippy runs with `-D warnings -W clippy::cast_possible_truncation`; there are no casts here.
6. Oracles for the Python surface (no API moves, so `__init__.pyi`, `tests/test_contract.py` frozen sets and `docs/compatibility.md` stay untouched): `tests/test_headers.py`, `tests/test_multivalue_headers.py` (key order, repeated keys -- #157, #12, #23), `tests/test_stdlib_parity.py` (its `DIVERGENCES` table must neither gain nor lose an entry: a new unexplained divergence or a `STALE` entry both fail), `tests/test_metadata_mode.py`, `tests/test_lazy_mode.py`, `tests/test_mime_tree.py`, `tests/test_tree_modes.py`, `tests/test_attachments.py` (disposition token), `tests/test_rfc_corpus.py`.
7. `fuzz/Cargo.toml` patches the same `../vendor/mailparse`, so `cargo fuzz run parse_agreement -- -max_total_time=300` is a free extra check that flat and tree parses still agree on headers.
8. `CHANGELOG.md` under `## [Unreleased]` / `### Changed`: one entry with the local before/after numbers, in the style of the 0.9.0 entry.
9. Do not add `memchr` to the root `Cargo.toml`; nothing in `src/` needs it for this change.

## Harness -- how to prove it

Build both sides once, then measure interleaved (the build cycle alone moves numbers ~16%, `.github/scripts/ab_median.py` docstring):

```bash
# base
git worktree add /tmp/fmp-base master
(cd /tmp/fmp-base && uv venv .venv && . .venv/bin/activate && uv pip install maturin pytest pytest-benchmark mail-parser && maturin develop --release)
# head (this branch)
uv venv .venv && . .venv/bin/activate && uv pip install maturin pytest pytest-benchmark mail-parser && maturin develop --release

# interleaved rounds, both sides, same machine, nothing else running
for r in 1 2 3 4 5; do
  /tmp/fmp-base/.venv/bin/pytest -q tests/benchmark --benchmark-min-rounds=25 --benchmark-json=base-$r.json
  .venv/bin/pytest -q tests/benchmark --benchmark-min-rounds=25 --benchmark-json=head-$r.json
done
```

Benchmarks expected to move (`tests/benchmark/test_read_message.py`):
- `test__fast_mail_parser___parse_metadata` (260-283) and `test__fast_mail_parser___parse_tree_metadata` (331): **expected -5 to -15%**. This is the primary number; if it moves less than the control noise floor the fast path is not earning its PATCH.md entry and the issue should be closed with the measurement.
- `test__fast_mail_parser___parse_many_metadata` (373-388, `threads=1`): same direction.
- `test__fast_mail_parser___parse_message`, `parse_tree`, `parse_lazy_*`: expected -1 to -3%, i.e. within noise; must not regress > 7%.
- `test__threaded___parse_many_small`, `test__threaded___parse_many_metadata_small` (227-232, 391-399): informational in the gate (`test.yml:466-470` passes `--informational test__threaded___`) because thread scheduling dominates. To see the effect, add single-threaded variants `test__fast_mail_parser___parse_many_small` and `test__fast_mail_parser___parse_many_metadata_small` calling `parse_many(batch, threads=1[, mode="metadata"])` on `[_small_message()] * SMALL_BATCH`, modelled on 157-165 and 373-388 (guarded with the same `pytest.skip` on `TypeError` so a base predating `mode=` does not take the gate down). Relates to round-1 "Benchmark gate only sees one 767 KiB base64 message: add gated small, QP and RFC 2047-heavy benchmarks" -- if that lands first, use its gated small benchmark instead of adding one. Expected: a few percent end to end (Python object construction per message is a large share there); report whatever is measured.
- Controls (`test__mail_parser___*`, `test__mailparser_lib___*`, `test__stdlib_email___*`) must stay flat; a treatment delta counts only when it clears their spread.

Correctness and lint surface, all required green:

```bash
cargo fmt --all -- --check
ruff check .
```

Put the local before/after table (machine, rounds, every benchmark, control noise floor) in the PR body. The Apple M4 is not a proxy for the x86 CI gate: this touches the per-header loop in a crate where code placement on the runners' Zen CPUs has swung results up to 96% (`CONTRIBUTING.md`, "Performance"), so the PR gate's interleaved A/B and, if it reports anything surprising, a `toolchain-ab.yml` run are the verdict. Re-run a large failure before acting on it.

## Acceptance criteria

- [ ] `vendor/mailparse/src/lib.rs`: `normalize_header` returns `chars.trim_start().to_owned()` when the value contains no `\n`, no `\r` and no `=?`, and otherwise delegates to the unchanged tokenizer path, kept as a separately named function.
- [ ] A differential test in the vendored crate compares the fast path against the tokenizer path over a generated corpus at every alignment plus the hand-written edge table above, and passes under `cargo test --manifest-path vendor/mailparse/Cargo.toml --target-dir /tmp/vendor-target`.
- [ ] `vendor/mailparse/PATCH.md` lists three changed functions, describes the third, and its sync procedure names it; the root `Cargo.toml` `[patch.crates-io]` comment no longer says "one function changed". `diff -r` against the registry 0.16.1 copy shows only `src/bytescan.rs`, `src/lib.rs`, `src/body.rs`, `Cargo.toml` and `PATCH.md`.
- [ ] `collect_headers` performs no `key.clone()`, keys its position map by `get_key_raw()`, pre-sizes both containers, and `tests/test_headers.py`, `tests/test_multivalue_headers.py`, `tests/test_stdlib_parity.py` pass with no change to `DIVERGENCES`.
- [ ] `disposition_token` tests presence with `get_first_header`, not `get_first_value`; `tests/test_attachments.py`, `tests/test_metadata_mode.py`, `tests/test_lazy_mode.py`, `tests/test_mime_tree.py` pass.
- [ ] No `memchr` (or any) dependency added to the root `Cargo.toml`; `Cargo.lock` unchanged apart from nothing.
- [ ] Interleaved local A/B (>= 5 rounds, controls reported): `test__fast_mail_parser___parse_metadata` improves by at least the control noise floor; no benchmark regresses more than 7%; the table is in the PR body.
- [ ] Single-threaded small-batch benchmarks (`threads=1`, full and metadata) exist -- added here or via the round-1 benchmark issue -- and their before/after is reported.
- [ ] The PR benchmark gate (x86 runner) passes; the run's CPU line is quoted in the PR.
- [ ] Full lint/test surface green: `pytest tests --ignore=tests/benchmark`, `cargo fmt --all -- --check`, `cargo clippy --all-targets -- -D warnings -W clippy::cast_possible_truncation`, `mypy --strict fast_mail_parser/`, `ruff check .`, vendored suite.
- [ ] `CHANGELOG.md` `[Unreleased]` entry with the measured numbers.

## Out of scope

- Rewriting `parse_header` (`vendor/mailparse/src/lib.rs:320-412`) on `memchr`, and `repair_missing_separator`'s `position` scan (`src/mail_parser.rs:275`): ~2.2 KB of header bytes per fixture parse, an expected 1-2 us against a 30 us metadata parse, with the highest edge-case surface in the crate (trailing-CR exclusion at 371-377, key-only lines at 346-360, `PreValue` skipping spaces but not tabs at 363-369) and a fourth hand-merged vendored function. Revisit only if a profile taken after this change shows `parse_header` above a few percent of metadata mode.
- Evaluating `get_body_encoded()` once per part (round-1 "Evaluate Body once per part").
- Reading Subject/Date/Content-ID out of the `collect_headers` table instead of `get_first_value` (round-1 "Share one envelope/part-classification helper pair"; see the #28 comment at `src/mail_parser.rs:1308-1311`).
- Making `tokenize_header_line` push into a caller-supplied `Vec` to save the per-line allocation on the folded/encoded slow path: a change to `header.rs`, which this issue deliberately leaves byte-identical to upstream.
- Any change to the Python-visible `headers` dict, its construction or caching (round-1 "Cache the headers dict per object").

## Related issues

- #231 — the Python dict cache above this
- #230
- #235
- #223 — `parse_qp_metadata` isolates `collect_headers`

kurok:
Closes #238.

Branched from `master` (`e7bbb05`). Independent of #245/#246/#247, but it edits `vendor/mailparse/PATCH.md` and the root `[patch.crates-io]` comment, which #245 also touches — see the rebase note.

## What changed

Every header value went through mailparse's two-stage tokenizer: a `Vec` per physical line, an outer `Vec`, a second `Vec` from the whitespace pass, and a result `String` grown with no capacity hint. **About six allocations to hand back the line it was given.**

That machinery is for folded values and RFC 2047 encoded words, and the common case is neither. On the 767 KiB fixture the root block has 24 headers, 8 continuation lines and exactly one encoded word; every part-level `Content-Type`, `Content-Transfer-Encoding`, `Content-Disposition`, `Content-ID`, `Date` and `MIME-Version` is one line with no `=?`. And the tokenizer is paid *more than once per header* — `collect_headers` reads them all, then mailparse re-tokenises `Content-Type` per part, `Content-Transfer-Encoding` on every `get_body_encoded()`, `Content-Disposition` on every `get_content_disposition()`, and the three flat parsers add `Subject`, `Date` and `Content-ID`. Roughly **sixty `get_value` calls per full parse**, ~55 of them for values the fast path answers with one `to_owned()`.

Three changes, no new dependency, `Cargo.lock` untouched, no output change:

1. **`normalize_header`** (vendored) returns `chars.trim_start().to_owned()` when the value has no `\n`, no `\r` and no `=?`; otherwise it delegates to the unchanged tokenizer, kept as `normalize_header_tokens`.
2. **`collect_headers`** keys its position map by `get_key_raw()` instead of a second owned `String`, and sizes both containers from `part.headers.len()`. Latin-1 decoding is injective, so raw-byte equality *is* the `String` equality it had — case-sensitive grouping included.
3. **`disposition_token`** tests presence with `get_first_header` instead of `get_first_value`, which was building and dropping a normalised `String` on the next line.

## Why the fast path is exactly equivalent

Not "close enough" — the comment and the test both make the argument:

`tokenize_header` runs `value.lines().map(str::trim_start)`, so with no `\n` that yields this string once (and nothing for `""`, where `trim_start` is also empty). A value produced by `parse_header` never ends in `\r` (only a non-CR non-LF byte advances `ix_value_end`), so the line is the whole value. `tokenize_header_line` with no `=?` pushes exactly one `maybe_whitespace(line)` token. `normalize_header_whitespace` passes a lone `Text` or `Whitespace` through unchanged.

`\r` is checked anyway — redundant for values `parse_header` produces, but it keeps the fast path correct for a `MailHeader` built any other way.

`fast_path_matches_the_tokenizer` asserts it regardless, against `normalize_header_tokens` as oracle, over a hand-written edge table (empty values, lone `\r`, `"=?"` split across the boundary, leading tabs, `"= ?"`, `"?="`) **and** a generated corpus at every length to 48 over an alphabet built from the bytes the check looks for.

## Numbers

Apple M4 (10 vCPU), rustc 1.98.0, CPython 3.12, 5 interleaved rounds. Pure-Python controls within **1.3%**.

|---|---|---|---|
| `parse_many_small_threads1` *(new)* | 4.056 ms | **3.133 ms** | **−23%** |
| `parse_many_metadata_small_threads1` *(new)* | 3.620 ms | **2.754 ms** | **−24%** |
| `parse_metadata` (767 KiB) | 0.037 ms | 0.027 ms | see below |
| `parse_message` (767 KiB) | 0.164 ms | 0.154 ms | −6.4% |

**The two `threads=1` batches are the reading that matters**, and they are new here (the issue asks for them). Serial, so no scheduler; small messages with no attachments, so no transfer decode — very nearly pure header work, which is exactly what this changes. ~23–24% off both.

**On the 767 KiB rows, a caveat I want to be explicit about.** `parse_metadata` reads 0.037 → 0.027 ms, which looks like 36%. It is not: master's 0.037 includes the M4 code-placement artifact documented on #244 — that benchmark was 0.030 ms before base64-simd was linked, and x86 has measured it flat on #244, #245, #246 and #247. Against the artifact-free 0.030 the real gain is ~10%. The x86 gate is the number to quote for that row.

| Benchmark | headers-fast | master | Delta | |
|---|---|---|---|---|
| `test__fast_mail_parser___full_read` | 0.161 ms | 0.171 ms | +6.0% |  |
| `test__fast_mail_parser___parse_lazy_all_attachments` | 0.172 ms | 0.181 ms | +4.9% |  |
| `test__fast_mail_parser___parse_lazy_untouched` | 0.040 ms | 0.050 ms | +25.1% |  |
| `test__fast_mail_parser___parse_many` | 1.309 ms | 1.379 ms | +5.3% |  |
| `test__fast_mail_parser___parse_many_metadata` | 0.215 ms | 0.294 ms | +37.0% |  |
| `test__fast_mail_parser___parse_message` | 0.154 ms | 0.164 ms | +6.4% |  |
| `test__fast_mail_parser___parse_message_strict` | 0.154 ms | 0.163 ms | +6.0% |  |
| `test__fast_mail_parser___parse_metadata` | 0.027 ms | 0.037 ms | +36.5% |  |
| `test__fast_mail_parser___parse_metadata_str` | 0.027 ms | 0.037 ms | +36.1% |  |
| `test__fast_mail_parser___parse_tree` | 0.157 ms | 0.168 ms | +7.4% |  |
| `test__fast_mail_parser___parse_tree_lazy_untouched` | 0.038 ms | 0.049 ms | +28.3% |  |
| `test__fast_mail_parser___parse_tree_metadata` | 0.027 ms | 0.038 ms | +38.4% |  |
| `test__mail_parser___parse_message` | 5.562 ms | 5.492 ms | -1.3% | control |
| `test__mailparser_lib___full_read` | 5.897 ms | 5.940 ms | +0.7% | control |
| `test__stdlib_email___full_read` | 7.536 ms | 7.477 ms | -0.8% | control |
| `test__threaded___parse_many` | 0.622 ms | 0.751 ms | +20.8% | informational |
| `test__threaded___parse_many_metadata_small` | 2.279 ms | 2.540 ms | +11.4% | informational |
| `test__threaded___parse_many_metadata_small_threads1` | 2.754 ms | 3.620 ms | +31.5% | informational |
| `test__threaded___parse_many_small` | 2.442 ms | 2.785 ms | +14.0% | informational |
| `test__threaded___parse_many_small_threads1` | 3.133 ms | 4.056 ms | +29.4% | informational |
| `test__threaded___threadpool_parse_email` | 0.620 ms | 0.646 ms | +4.2% | informational |
| `test__threaded___threadpool_parse_email_small` | 14.442 ms | 13.758 ms | -4.7% | informational |

Noise floor from the pure-Python controls: **1.3%** (they cannot be affected by the build, so this is measurement error).

</details>

## Checks run locally

`pytest tests --ignore=tests/benchmark` (811 passed, 3 skipped — including `test_headers.py`, `test_multivalue_headers.py`, `test_stdlib_parity.py` with no `DIVERGENCES` change, and the attachment/metadata/lazy/tree suites that exercise `disposition_token`), the vendored suite (54 + 25 passed) via `--target-dir` outside the tree, `cargo fmt` for **both** manifests, `cargo clippy --all-targets -- -D warnings -W clippy::cast_possible_truncation`, `mypy --strict`, `ruff check .` — all green. `git diff --stat master -- Cargo.lock` is empty.

The M4 is not a proxy for x86; the **Benchmark quality gate** is the verdict.

## Rebase note

#245 also edits `vendor/mailparse/PATCH.md` (making it "three functions changed" for the quoted-printable decoder) and the root `[patch.crates-io]` comment. Whichever of the two lands second needs those merged by hand — the conflicts are additive, but the "three functions" counts collide and want reading rather than taking one side. I have done this merge twice already this series and will handle it.


kurok:
Rebased onto `2aee3df` now that #245 has landed — the `PATCH.md` collision this PR's Rebase note predicted. Three files conflicted (`PATCH.md`, the root `[patch.crates-io]` comment, `CHANGELOG.md`); the vendored copy is now **four** changed functions, and the `diff -r` claim and sync checklist name `normalize_header`/`normalize_header_tokens` alongside `src/qp.rs`.

**The new baseline settles the caveat in the PR body.** I wrote that `parse_metadata`'s 0.037 → 0.027 was not really 36%, because master's 0.037 carried the M4 placement artifact and the artifact-free figure was 0.030, so the real gain was ~10%. On `2aee3df` master measures **0.030** and the gain measures **+10.4%** — which is the number that reasoning predicted, arrived at from a different base.

Re-measured, 5 interleaved rounds, pure-Python controls within 2.1%:

| Benchmark | master `2aee3df` | this branch | |
|---|---|---|---|
| `parse_many_metadata_small_threads1` | 3.671 ms | **2.746 ms** | **+33.7%** |
| `parse_many_small_threads1` | 4.056 ms | **3.171 ms** | **+27.9%** |
| `parse_qp_message_metadata` | 0.022 ms | **0.019 ms** | +18.2% |
| `parse_metadata` | 0.030 ms | **0.027 ms** | +10.4% |
| `parse_message` | 0.158 ms | 0.154 ms | +2.7% |
| `parse_qp_message` | 0.134 ms | 0.130 ms | +2.8% |

The two `threads=1` batches remain the reading that matters: serial, small messages, no transfer decode — very nearly pure header work.

Worth noting the quoted-printable rows, which did not exist when this branch was first cut: `parse_qp_message_metadata` is header-dominated on a message with 29 headers, and it moves 18%. #229 and #238 are independent changes to different parts of the same parse and they compose.

Re-verified after the rebase: 811 passed / 3 skipped, the vendored suite 58 + 25 (both `fast_path_matches_the_tokenizer` and all four of #229's QP differential tests), fmt for both manifests, clippy, mypy --strict, ruff.

kurok:
**The gate failed, and it was right to.** `test__fast_mail_parser___parse_qp_message` read **+9.2% slower** on x86 (0.246 → 0.268 ms), past the 7% threshold, while everything else on the same run improved:

| Benchmark | base | this revision | |
|---|---|---|---|
| `parse_qp_message` | 0.246 ms | 0.268 ms | **+9.2% slower** |
| `parse_tree_metadata` | 0.055 ms | 0.049 ms | −10.0% |
| `parse_many_metadata_small` | 4.165 ms | 3.739 ms | −10.2% |
| `parse_metadata` | 0.052 ms | 0.048 ms | −8.1% |
| `parse_many_small` | 4.705 ms | 4.298 ms | −8.7% |

Two things make me read that as placement rather than as this change:

1. **There is no mechanism.** Nothing here touches quoted-printable decoding. The QP path's only contact with this diff is `get_body_encoded()` → `get_first_value("Content-Transfer-Encoding")`, which the fast path makes *cheaper*.
2. **The two CPUs disagree in direction.** The M4 reads that same benchmark **3.2% faster**, over 5 interleaved rounds, twice. #204 is exactly this.

So: `normalize_header_tokens` is now marked `#[inline(never)]`. That is not a workaround dressed up — after the fast path it genuinely *is* the cold half, running only for folded values and encoded words, and leaving it inlinable put its two `Vec`s and its loop inside a function most callers leave after one branch.

**What I cannot tell you is whether it works.** The M4 does not reproduce the regression, so I have no local instrument for this: the branch measures the same before and after the attribute here. This is the principled change *and* the documented first thing to try, but the gate on x86 is the only thing that can decide it. If it does not clear, the next step is #240's salted layout A/B rather than another guess.

Everything else re-verified: 811 passed / 3 skipped, vendored suite 58 + 25, fmt both manifests, clippy, mypy --strict, ruff.

kurok:
**Third gate run made it worse (+12.9%), so I have stopped guessing and rescoped this PR instead.**

Against an unmoving 0.246 ms base, `parse_qp_message` measured across three binaries:

| attempt | change | `parse_qp_message` head | delta |
|---|---|---|---|
| 1 | as authored | 0.268 ms | +9.2% |
| 2 | `#[inline(never)]` on the now-cold tokenizer path | 0.265 ms | +7.5% |
| 3 | also `#[inline(never)]` on `qp::decode_robust` | 0.277 ms | **+12.9%** |

On all three runs `parse_qp_message_metadata` — the same message, same headers, same structure, decode skipped — got **6.5% faster**, and every metadata and small-batch benchmark improved 6–10%. The M4 reads `parse_qp_message` ~3% *faster* throughout and cannot see any of this. Three binaries is enough to say that guessing at placement does not converge.

**What this PR is now.** The two changes that need no vendored edit:

- `collect_headers` keys its position map by raw key bytes instead of a second owned `String`, and pre-sizes both containers.
- `disposition_token` uses `get_first_header` instead of building and dropping a normalised value.

|---|---|---|---|
| `parse_many_small_threads1` *(new)* | 4.048 ms | **3.839 ms** | −5.2% |
| `parse_many_metadata_small_threads1` *(new)* | 3.622 ms | **3.442 ms** | −5.0% |
| `parse_qp_message_metadata` | 0.022 ms | 0.021 ms | −6.0% |

Controls within 1.7%. `vendor/` is untouched — `git diff origin/master -- vendor/` is empty — so `PATCH.md`, the root patch comment and the vendored suite are all back to master's state.

**What came out, and what should happen to it.** The `normalize_header` fast path was worth far more: the same two benchmarks read **−28% to −34%** with it. It is a real ~25-point win sitting behind a layout problem that this repo already has an issue for. #240 is exactly that instrument — a salted layout A/B on x86 — and the fast path should go back through the gate behind it rather than behind a fourth guess. I have not filed anything; say the word and I will open an issue carrying the three measurements, the equivalence argument and the differential test, so none of that work is lost.

I would rather land 5% that is understood than 30% that fails the gate for a reason I cannot name.

PR #253, opened 2026-09-17
Evaluate a part's body once, and stop copying text bodies twice (#230)

### LINKED ISSUE #230: Evaluate Body once per part: drop the redundant text-body copy, the second Content-Transfer-Encoding lookup, and the scalar '=' scan on quoted-printable parts
## Problem

For every non-multipart part, the full and lazy parsers do the transfer-encoding dispatch twice and copy text bodies once more than necessary before the bytes reach Python. The per-part sequence today is: `part.get_body_raw()?` (which calls `get_body_encoded()` -> header lookup + `to_lowercase()` String, then for 7bit/8bit/binary parts copies the body slice into a fresh `Vec<u8>`), then `quoted_printable_invalid_escape(&part)` (which calls `get_body_encoded()` again and, for quoted-printable parts, walks the whole encoded body byte-at-a-time), then `decode_charset(&content, ..)` which for a valid UTF-8 body gets a `Cow::Borrowed` back from encoding_rs and immediately `into_owned()`s it -- a second body-sized copy, even though the caller already owns a `Vec<u8>` that `String::from_utf8` could take over without copying.

None of this is visible on the CI gate fixture: `tests/data/large_message.eml` has no quoted-printable part and its two text parts are under 10 KB, so the extra copy and the second header lookup are noise there. Real corpora (newsletters, transactional HTML) are dominated by 100-500 KB `text/html` bodies in 8bit or quoted-printable, where a whole-body memcpy and a whole-body scalar scan are real work -- and the scan is exactly the loop shape CONTRIBUTING.md documents as placement-sensitive on the Zen runners. The maintainability half is independent of the numbers: `get_body_encoded()` should be evaluated once per part and its `Body` threaded to the helpers, instead of three helpers each re-deriving it.

Honest scope: on a quoted-printable body the dominant cost is the `quoted_printable` crate's own `_decode` (a `filter_map(..).collect::<String>()` pass followed by a per-byte push loop), which this issue does not touch. This issue removes the avoidable work *around* the decoders; it is a low-to-mid single-digit % win on large 8bit/QP text bodies plus a cleanup, not a headline number.

## Evidence

- `src/mail_parser.rs:1375` (full mode, `Mail::from_payload` per-part loop): `let content = part.get_body_raw()?;` -- then `:1379`: `let lossy_escape = quoted_printable_invalid_escape(&part);` -- two `get_body_encoded()` evaluations per part.
- `src/mail_parser.rs:1404` and `:1416`: `let (text, fell_back) = decode_charset(&content, &part.ctype);` -- `content` (the Vec) is read once and dropped.
- `src/mail_parser.rs:818` (lazy mode): `let lossy_escape = quoted_printable_invalid_escape(&part);` and `:846-847`: `let content = part.get_body_raw()?; let (text, fell_back) = decode_charset(&content, &part.ctype);` -- same double dispatch.
- `vendor/mailparse/src/lib.rs:776-782`:
  ```rust
  pub fn get_body_raw(&self) -> Result<Vec<u8>, MailParseError> {
      match self.get_body_encoded() {
          Body::Base64(body) | Body::QuotedPrintable(body) => body.get_decoded(),
          Body::SevenBit(body) | Body::EightBit(body) => Ok(Vec::<u8>::from(body.get_raw())),
          Body::Binary(body) => Ok(Vec::<u8>::from(body.get_raw())),
      }
  }
  ```
  For 7bit/8bit/binary this is a pure copy of the body slice.
- `vendor/mailparse/src/lib.rs:825-832`: `get_body_encoded` does `self.headers.get_first_value("Content-Transfer-Encoding").map(|s| s.to_lowercase())` -- a linear header scan and a String allocation on every call.
- `src/mail_parser.rs:433-439`:
  ```rust
  fn decode_charset(body: &[u8], ctype: &ParsedContentType) -> (String, bool) {
      if let Some(charset) = Charset::for_label(ctype.charset.as_bytes()) {
          (charset.decode(body).0.into_owned(), false)
      } else {
          (decode_ascii(body).into_owned(), true)
      }
  }
  ```
  `charset-0.1.5/src/lib.rs:318-325` (`Charset::decode`) BOM-sniffs with `Charset::for_bom(bytes)` and then calls `decode_without_bom_handling`; `encoding_rs-0.8.35/src/lib.rs:3104-3116` returns `Cow::Borrowed` when `utf8_valid_up_to(bytes) == bytes.len()`. So for the common case (UTF-8 label, valid UTF-8, no BOM) `into_owned()` is a body-sized memcpy of bytes the caller already owns.
- `src/mail_parser.rs:103-131`:
  ```rust
  #[inline(never)]
  fn quoted_printable_invalid_escape(part: &ParsedMail<'_>) -> Option<usize> {
      let Body::QuotedPrintable(body) = part.get_body_encoded() else { return None; };
      let raw = body.get_raw();
      let mut at = 0;
      while at < raw.len() {
          if raw[at] != b'=' { at += 1; continue; }
          ...
  ```
  A scalar byte loop over the whole encoded QP body, in addition to the decoder's own passes. `CONTRIBUTING.md:315-345` records that scalar loops of this shape straddling a 64-byte boundary ran at half speed on the Zen runners (+96% swings), while an M4 does not show it.
- `tests/data/large_message.eml:39-40` (`text/plain; charset=utf-8` / `8bit`), `:56-57` (`text/HTML; charset=utf-8` / `base64`), `:145-149` and `:3946-3950` (base64 attachments) -- no quoted-printable part, text parts small. Known profile of a full parse of this fixture (M4, 0.225 ms): 53% `data_encoding::decode_base_mut`, 27% base64 whitespace strip, 7.5% memmem, everything else <5%. This issue's work lives in that <5%; the gate cannot see it.
- `Cargo.toml:37-43`: root `[dependencies]` are `charset`, `mailparse`, `pyo3` only; `Cargo.lock:74-75` already contains `memchr 2.8.3` via `vendor/mailparse/Cargo.toml` (`[dependencies.memchr] version = "2.7.0"`). `fuzz/Cargo.toml:22-25` lists `libfuzzer-sys`, `charset`, `mailparse`, and its comment at `:6-9` states the core "uses only `charset`, `mailparse` and `std`"; `fuzz/fuzz_targets/parse_agreement.rs:48-49` and `parse_email.rs:22-23` include `src/mail_parser.rs` by `#[path]`.
- `quoted_printable-0.5.1/src/lib.rs:202-283` (`_decode`): `input.into_iter().filter_map(..).collect::<lib::String>()` then `filtered.lines()` with a per-byte `decoded.push(..)` loop -- the decoder itself is the bulk of a QP parse and is out of scope here (separate issue if wanted).

## Proposal

In `Mail::from_payload` (full) and `lazy_from_payload` (lazy), evaluate `let body = part.get_body_encoded();` once per part and thread `&body` to everything that needs it. Concretely, in `src/mail_parser.rs`:

1. `quoted_printable_invalid_escape(body: &Body<'_>) -> Option<usize>` (was `&ParsedMail`). Keep `#[inline(never)]`. Drive the scan with `memchr::memchr_iter(b'=', raw)`; the predicate per hit is unchanged, and the result is provably identical because the bytes following a valid escape (two hex digits) or a soft break (CR LF / LF) are never `=`, so visiting every `=` in order and returning the first that fails the check equals the current stepping loop:
   ```rust
   memchr::memchr_iter(b'=', raw).find(|&at| {
       let rest = &raw[at + 1..];
       !(rest.starts_with(b"\r\n")
           || rest.starts_with(b"\n")
           || (rest.len() >= 2 && rest[0].is_ascii_hexdigit() && rest[1].is_ascii_hexdigit()))
   })
   ```
2. `encoded_size(body: &Body<'_>) -> usize` (was `&ParsedMail`, `:528-534`) so lazy attachments reuse the same `Body`.
3. A `fn body_to_vec(body: &Body<'_>) -> Result<Vec<u8>, MailParseError>` that is `get_body_raw`'s match verbatim, used for attachments (their `content: Vec<u8>` stays byte-identical).
4. For bodies, branch on the variant:
   - `SevenBit | EightBit | Binary` -> `decode_charset(b.get_raw(), &part.ctype)` on the borrowed slice. Removes the `Vec` copy; the `into_owned()` stays because `Mail.text_plain: Vec<String>` (`src/fast_mail_parser.rs:793`) owns its strings.
   - `Base64 | QuotedPrintable` -> `decode_charset_owned(b.get_decoded()?, &part.ctype)`, new:
   ```rust
   /// `decode_charset` for a body the caller owns: when the label resolves to
   /// UTF-8 and the bytes carry no BOM, hand the Vec to `String::from_utf8`
   /// instead of validating with encoding_rs and then copying. Same output:
   /// encoding_rs returns the input unchanged for valid UTF-8, and the two
   /// cases where it would not (BOM present, invalid sequences) take the old
   /// path. No `unsafe`, per the module invariant.
   fn decode_charset_owned(body: Vec<u8>, ctype: &ParsedContentType) -> (String, bool) {
       let Some(charset) = Charset::for_label(ctype.charset.as_bytes()) else {
           return (decode_ascii(&body).into_owned(), true);
       };
       if charset.name() == "UTF-8" && Charset::for_bom(&body).is_none() {
           match String::from_utf8(body) {
               Ok(text) => return (text, false),
               Err(err) => {
                   let body = err.into_bytes();
                   return (charset.decode(&body).0.into_owned(), false);
               }
           }
       }
       (charset.decode(&body).0.into_owned(), false)
   }
   ```
   `Charset` is `PartialEq + Copy` and exposes `name()`/`for_bom()` (charset-0.1.5 `src/lib.rs:176-177`, `:268-274`, `:279`), so no `encoding_rs` direct dependency is needed. Note `Charset::for_label` maps `utf8`, `unicode-1-1-utf-8` etc. to the same charset, so compare the resolved charset, never the raw label.

The `fell_back` flag, `warn_charset`, `warn_transfer_decode`, warning order and `part_path` indices are unchanged. Attachments keep exactly `get_body_raw`'s bytes.

New dependency: `memchr = "2.7"` added to the root `Cargo.toml` `[dependencies]` **and** to `fuzz/Cargo.toml` `[dependencies]`. It adds no crate to `Cargo.lock` (already present at 2.8.3 through `vendor/mailparse`), is MIT OR Unlicense with no dependencies of its own, and is the same crate the maintainer chose for the boundary search in #213.

1. `src/mail_parser.rs`: add `use memchr::memchr_iter;` (or fully qualified). Change the signatures of `quoted_printable_invalid_escape` (`:103-131`) and `encoded_size` (`:528-534`) to take `&Body<'_>`; add `body_to_vec` and `decode_charset_owned` next to `decode_charset` (`:433-439`). Update the doc comments: `decode_charset`'s promise "mirrors mailparse's internal `get_body_as_string` exactly" (`:420-432`) now also has to state why `decode_charset_owned` is equivalent (BOM and invalid-UTF-8 cases fall through to `charset.decode`).
2. Restructure the per-part body handling in `Mail::from_payload` (`:1360-1430`) and `lazy_from_payload` (`:800-870`): one `let body = part.get_body_encoded();`, `lossy_escape` from `&body`, attachments via `body_to_vec(&body)?`, text bodies via the variant match above. Keep the `?` on decode failure *before* any warning is pushed, as today. Keep the `#[inline(never)]`/`#[cold]` conventions (`:95-104`, `:151-158`); do not add generic helpers that the per-part loop would instantiate more than once (`CONTRIBUTING.md:311-318` on `catch_panics`).
3. `Cargo.toml` (`:37-43`): add `memchr = "2.7"` with a one-line comment (already in the graph via the vendored copy). `fuzz/Cargo.toml` (`:22-25`): add the same line and rewrite the comment at `:6-9` ("uses only `charset`, `mailparse` and `std`") to include `memchr`. `CONTRIBUTING.md:199-200` says `fuzz/Cargo.toml` duplicates the `charset` and `mailparse` versions -- add `memchr` to that sentence. `cargo update` is not needed; verify `git diff Cargo.lock` shows only the new edge under `fast_mail_parser`, no new package.
4. `vendor/mailparse/` is **not** touched by this issue. If you find yourself editing it, stop: `PATCH.md` would need updating and the sdist include glob (`pyproject.toml:46-48`, `vendor/mailparse/**/*`) sweeps a stray `target/` in -- never run cargo inside `vendor/mailparse` without `--target-dir` elsewhere.
5. Oracles (must be green before and after): `tests/test_parse_warnings.py:528-600` (`transfer-decode-lossy`: invalid `=ZZ`, valid `=3D` and soft breaks, bare-LF not reported, attachment `part_path`, strict rejects); `tests/test_lazy_mode.py:109` and `:137` (bodies and the warning list identical to full mode over the RFC corpus); `tests/test_stdlib_parity.py` (its `DIVERGENCES` table must not need a new row -- if it does, the change altered output); the fuzz `parse_agreement` target (invariant 7: lazy bodies and warnings equal the full parse) -- run it for a few minutes locally with `RUSTUP_TOOLCHAIN=nightly cargo fuzz run parse_agreement -- -max_total_time=300` if nightly is available.
6. Add tests **before** changing the decode path, and confirm they pass on master, so they pin current behaviour rather than the new code:
   - `tests/test_parse_warnings.py` or a new `tests/test_charset_decode.py`: (a) an 8bit `text/plain; charset=utf-8` body starting with `\xef\xbb\xbf` decodes with the BOM stripped (`text_plain[0]` has no `﻿`); (b) a base64 body labelled `utf-8` whose decoded bytes start with `\xff\xfe` (UTF-16LE BOM) decodes as UTF-16 (`"hi"` from `b"\xff\xfeh\x00i\x00"`); (c) a base64 body labelled `utf-8` containing an invalid sequence (`b"ok \xff ok"`) yields `"ok � ok"` with `warnings == []`; (d) `charset=utf8` and `charset=UTF-8` spellings behave the same as `utf-8`. These are the cases where the new fast path must fall back.
   - A QP body whose `=` density is high (`=3D` every few bytes) and one whose only `=` is a trailing `=` at end of body, both asserted against the current warning output (the trailing `=` is reported at the last index).
7. Add the benchmark that makes the work visible, in `tests/benchmark/test_read_message.py`, generated in-module like `_small_message()` (`:215-224`) so no new fixture file is committed:
   - `_html_message(cte: str) -> bytes` building a ~300 KB `text/html; charset=utf-8` single-part message with `email.message.EmailMessage.set_content(html, subtype="html", charset="utf-8", cte=cte)`, HTML containing attributes (`=`), non-ASCII text, and 76-column wrapping so QP produces soft breaks; deterministic content.
   - `test__fast_mail_parser___parse_html_qp` (`cte="quoted-printable"`) and `test__fast_mail_parser___parse_html_8bit` (`cte="8bit"`), full mode; and `test__fast_mail_parser___parse_html_qp_lazy` (`mode="lazy"`, guarded with the same `pytest.skip` pattern as `:283-304` so a base without `mode=` still runs the gate). Assert once outside the timed call that `mail.text_html[0]` is non-empty and `mail.warnings == []`.
8. `CHANGELOG.md` under `## [Unreleased]` / `### Changed`: one entry naming the removed copy, the single `Body` evaluation, the `memchr` scan, and the measured numbers with the machine.

## Harness -- how to prove it

Build both sides once, then interleave measurements (the pattern `.github/workflows/test.yml:449-456` uses; a one-shot A/B on this project reads ~16% apart on a no-op change, `ab_median.py:8-12`):

```bash
# two venvs, one per side; base = master 73f56da, head = your branch
uv venv .venv-base && uv venv .venv-head
for side in base head; do
  ( source .venv-$side/bin/activate && uv pip install maturin pytest pytest-benchmark mail-parser )
done
git worktree add ../fmp-base 73f56da
( source .venv-base/bin/activate && cd ../fmp-base && maturin develop --release )
( source .venv-head/bin/activate && maturin develop --release )
# interleaved rounds, head/base alternating; the new benchmarks skip on base only where the API is missing
for round in 1 2 3; do
  for side in head base; do
    .venv-$side/bin/pytest -q tests/benchmark --benchmark-min-rounds=25 --benchmark-json=$side-$round.json
  done
done
```

Note the new benchmarks exist only on the head side unless you also add them to the base worktree's `tests/benchmark/test_read_message.py` (copy the file across; it is pure Python and measures the installed extension), otherwise `ab_median.py` drops them as "missing from some rounds". Do copy it -- those three rows are the point of the exercise.

Expected outcome:
- No treatment benchmark regresses by more than 7% (the gate threshold). `test__fast_mail_parser___parse_message`, `..._parse_message_strict`, `..._parse_many`, `..._parse_tree`, `..._parse_metadata`, `..._parse_lazy_*`, `..._parse_tree_*` are expected flat (within the control spread printed by `ab_median.py`).
- `test__fast_mail_parser___parse_html_8bit`: expected -2 to -6% (one 300 KB alloc+copy and one header lookup + String removed).
- `test__fast_mail_parser___parse_html_qp` and `..._parse_html_qp_lazy`: expected -3 to -10% (the scalar scan vectorised, one copy removed); the remaining time is the `quoted_printable` decoder, so do not expect more.
- `test__threaded___parse_many_small`: informational; expected flat to slightly better.
- Put the before/after table (with the CPU line `ab_median.py` prints) in the PR body. The M4 is not a proxy for x86: the scan being replaced is the Zen-placement-sensitive shape, so confirmation is the PR gate on the EPYC runner (`test.yml`) and, if the gate reads within noise, a `toolchain-ab.yml` run against the branch.

Full local CI surface, all must pass:

```bash
cargo fmt --all -- --check
ruff check .
# fuzz harness still builds with the new dependency (nightly):
( cd fuzz && RUSTUP_TOOLCHAIN=nightly cargo build --release )
```

## Acceptance criteria

- [ ] `get_body_encoded()` is called exactly once per non-multipart part in both `Mail::from_payload` and `lazy_from_payload` (grep shows `quoted_printable_invalid_escape` and `encoded_size` take `&Body<'_>`).
- [ ] 7bit/8bit/binary text bodies are charset-decoded from the borrowed slice; no `Vec<u8>` is built for them.
- [ ] Base64/QP text bodies labelled UTF-8 without a BOM reach `String` via `String::from_utf8(vec)` with no intermediate copy; BOM-prefixed and invalid-UTF-8 bodies produce byte-identical output to master (new tests in step 6 pass on master and on the branch).
- [ ] `quoted_printable_invalid_escape` uses `memchr::memchr_iter`; `tests/test_parse_warnings.py:528-600` and the new high-density / trailing-`=` tests pass unchanged.
- [ ] Attachment `content` is byte-identical to master for every fixture (`tests/test_lazy_mode.py:87`, parity suite, no new `DIVERGENCES` row).
- [ ] `memchr` is a direct dependency in both `Cargo.toml` and `fuzz/Cargo.toml`; `git diff Cargo.lock` adds no new package; `fuzz/Cargo.toml:6-9` and `CONTRIBUTING.md:199-200` name it; `cargo build` in `fuzz/` succeeds.
- [ ] `vendor/mailparse/` unchanged (`git diff --stat vendor/` empty).
- [ ] Three new benchmarks exist and run on both sides of the A/B; the interleaved A/B (3 rounds, `AB_THRESHOLD_PCT=7`) shows no treatment regression and the numbers are in the PR body with the CPU named.
- [ ] The PR gate on the x86 runner is green.
- [ ] Full lint/test surface above passes; `CHANGELOG.md` has the entry.

## Out of scope

- Replacing or vendoring the `quoted_printable` crate's `_decode` (the actual dominant cost of a QP part) -- a separate issue with its own A/B.
- Making `Mail`/`LazyMail` hold borrowed or `Cow` text to avoid the last copy before the PyO3 `str` conversion (touches the binding layer and the `parse_many` result ownership).
- The MIME tree API (`MimePart::build` `:935-943`, `build_node` `:1095`), which returns bytes and has no charset step; its `get_body_raw` calls are single and stay as they are.
- Any change to `vendor/mailparse` or to the warning taxonomy.

## Related issues

- #229
- #223
- #238
- #234 — the envelope/classify helpers this per-part change sits next to

kurok:
Closes #230 — with one deliberate omission, measured. See **What I took back out**.

Branched from `master` (`5027ca5`).

## What changed

`get_body_encoded()` re-reads a part's headers to find its transfer encoding, and both the full and lazy parsers called it **two or three times per part** — for the quoted-printable escape check, for the encoded size, and again inside `get_body_raw`. It is now called once and threaded to all three.

Two copies went with it, and neither bought anything:

- A **7bit/8bit/binary** body *is* its raw bytes, so `get_body_raw` copied it into a `Vec` only so `decode_charset` could borrow it straight back.
- A **base64 or quoted-printable** body, once transfer-decoded, was lent to `encoding_rs` by reference — which validates UTF-8 and hands back a `Cow::Borrowed` — and `into_owned()` then copied the whole body a second time.

So text bodies now decode from the borrowed slice where the encoding is already plaintext, and give the decoded `Vec` to `String::from_utf8` where it is not.

## Why the shortcut is safe

The two cases where `String::from_utf8` is *not* equivalent to `encoding_rs` are routed to the original path: a **BOM**, which `encoding_rs` strips and `from_utf8` would keep as U+FEFF, and **invalid sequences**, which `encoding_rs` replaces with U+FFFD and `from_utf8` rejects outright. The resolved charset is compared, never the raw label — `Charset::for_label` maps `utf8`, `UTF-8` and `unicode-1-1-utf-8` onto one charset.

`tests/test_charset_decoding.py` pins all of it: BOMs across four transfer encodings, invalid UTF-8, four spellings of the label, a non-UTF-8 charset, and an unknown one that must still warn.

**All sixteen pass against master as well.** That is the point — they are the oracle for the claim that the shortcut changed nothing, not a test of the shortcut.

## What I took back out

The issue also asks for the escape scan to use `memchr_iter`. I implemented it, benchmarked it, and removed it.

The new `parse_qp_dense_escapes` benchmark — one `=` every three bytes, the scan's worst case — came back **10.1% slower** with `memchr`, against a 1.2% control floor. `memchr`'s per-call setup never amortises when hits are three bytes apart, while the byte loop's test is a single comparison. Reverting **that hunk alone** moved it to −1.7% and left every other gain in place, which is how I know it was the `memchr` and not code placement:

| Benchmark | with `memchr` | without |
|---|---|---|
| `parse_qp_dense_escapes` | **−10.1%** | −1.7% |
| `parse_many_small_serial` | +6.9% | **+8.6%** |
| `parse_qp_message` | +12.5% | +2.4% |

The dependency declaration came out with it, so `Cargo.toml`, `fuzz/Cargo.toml` and `Cargo.lock` are unchanged from master.

(The `parse_qp_message` figure moved a lot between those two runs — 12.5% against 2.4% — which given this repo's current ±15% placement scatter I would not read as meaning anything. The two numbers I do trust are the dense-escape one, because it has a mechanism and reverting one hunk moved it back, and `parse_many_small_serial`, because it is consistent across both runs.)

## Numbers

Apple M4 (10 vCPU), 5 interleaved rounds, pure-Python controls within **0.5%**:

|---|---|---|---|
| `parse_many_small_serial` | 4.054 ms | **3.733 ms** | −8% |
| `parse_8bit_text` *(new)* | 0.003 ms | 0.003 ms | −8% |
| `parse_qp_message` | 0.135 ms | 0.131 ms | −2.4% |
| `parse_qp_dense_escapes` *(new)* | 0.017 ms | 0.017 ms | flat |

Three benchmarks are added: a plain 8bit body (almost entirely the removed copy plus charset validation), a base64 UTF-8 body (the owned-decode path), and the dense-escape worst case that caught the `memchr` regression.

## Checks run locally

`pytest tests --ignore=tests/benchmark` (843 passed, 3 skipped), the full benchmark suite (32 passed), `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings -W clippy::cast_possible_truncation`, `mypy --strict`, `ruff check .`. `git diff --stat origin/master -- vendor/ Cargo.lock` is empty. Attachment `content` is byte-identical — `body_to_vec` is `get_body_raw`'s match verbatim.


kurok:
**The gate was right, and fixing it properly found a second problem.**

The failure was `parse_base64_utf8_text` at +14.5%. That benchmark ran in **11 µs** on the runner — an order of magnitude below every other gated one (`parse_message` 317 µs, `parse_metadata` 50 µs) — so a 2 µs wobble *is* a 14.5% verdict, and most of what it measured was fixed per-call cost rather than the body handling it is named for. All three benchmarks this branch adds are now ~10x larger: **21 / 49 / 175 µs** on an M4.

At that size it said something the small one could not.

**The owned-decode shortcut is a pessimisation.** Handing the decoded `Vec` to `String::from_utf8` instead of lending it to `encoding_rs` genuinely saves a copy — and measured **8% slower** on a 128 KB UTF-8 body, 0.045 → 0.049 ms with *zero* variance across five rounds on both sides. `encoding_rs` validates UTF-8 with SIMD; `std` does not, and that gap is larger than the copy it saves.

Reverting that one arm:

| Benchmark | with the shortcut | without |
|---|---|---|
| `parse_base64_utf8_text` | **−8.0%** | **+0.3%** |
| `parse_qp_message` | +10.4% | **+9.8%** |
| `parse_8bit_text` | +7.8% | **+6.4%** |

So base64 and quoted-printable keep their original route, `decode_charset_owned` is gone, and `decode_body`'s comment records the measurement so the next person does not re-derive it.

**That is the second thing #230 prescribed that measurement refused**, after the `memchr` escape scan (10% slower on its worst case). What survives is what the issue got right, and it is still worth having:

| Benchmark | base | this branch | |
|---|---|---|---|
| `parse_qp_message` | 0.143 ms | **0.131 ms** | −9.8% |
| `parse_8bit_text` | 0.022 ms | **0.021 ms** | −6.4% |
| `parse_base64_utf8_text` | 0.045 ms | 0.045 ms | flat |
| `parse_qp_dense_escapes` | 0.183 ms | 0.182 ms | flat |

Controls within 2.4%.

`tests/test_charset_decoding.py` stays. It was written as the oracle for the shortcut, but it pins BOM handling, invalid UTF-8 and charset-label aliasing — all of which pass on master too — so it is a good regression pin regardless.

**One note on the instrument.** I tried to dispatch `layout-ab.yml` from this branch to get the failing benchmark's own layout spread, and could not: the branch predates #249, so the workflow is not on it. Worth knowing for the next person who wants per-branch layout numbers — the branch has to be rebased past #249 first.

Issue #236, opened 2026-08-30
Make the PyO3-free core a workspace lib crate instead of a `#[path]`-included file, and give it its first `cargo test`


kurok:
## Problem

The parsing core `src/mail_parser.rs` is written to be PyO3-free -- its module doc promises it "can be exercised and unit-tested independently of any Python runtime" -- but nothing can actually *link* it. The only Cargo package in the tree is a `cdylib` whose lib root is the PyO3 binding layer, so the core is reachable only as a `mod` inside that crate. Every other consumer has to include the source file textually: both fuzz targets do `#[path = "../../src/mail_parser.rs"] mod mail_parser;`, which in turn forces `fuzz/Cargo.toml` to re-declare `charset`, `mailparse` and the `[patch.crates-io]` entry and keep them in lockstep with the root manifest by hand (PATCH.md carries two steps that exist only for that).

The practical cost today: there is not a single `#[test]` in `src/` -- `cargo test` cannot link a test binary for a `cdylib` built with `pyo3/extension-module` -- so the only Rust-level check on the core is the nightly fuzz harness. Regressions in pure functions like `repair_missing_separator`, `quoted_printable_invalid_escape`, `collect_headers` ordering or the `parse_date_epoch` `Ok(0)` guard can only be caught through Python. The `pub(crate)` surface of the core is also accidental (whatever the textual include happens to expose) rather than a designed API, and any future split of the core into submodules would have to survive being `#[path]`-included from two places.

Speed is this library's reason to exist, and this change claims **no** speed win: the same code lands in the same `cdylib` under `lto = true`, `codegen-units = 1`. What it buys is the ability to test the hot paths in Rust, on stable, without Python -- which every later performance PR (base64/QP decoders, body evaluation, parse_many scheduling) needs as an oracle that is cheaper than the full pytest + benchmark cycle.

## Evidence

All at master `73f56da`.

- `Cargo.toml:32-35` -- the only crate, binding as lib root, `cdylib` only:
  ```toml
  [lib]
  name = "fast_mail_parser"
  path = "src/fast_mail_parser.rs"
  crate-type = ["cdylib"]
  ```
[...]

PR #256, opened 2026-09-17
Measure the core directly: a Rust bench crate with allocation counting (#225)

### LINKED ISSUE #225: Add a std-only Rust bench crate (timing + counting allocator) and document a symbolisable release build
## Problem

The only way to time this library today is a wheel build followed by `pytest tests/benchmark` through PyO3, and the only metric is wall time. After #213/#214/#218 the remaining ~80% of a full parse of `tests/data/large_message.eml` (0.225 ms on an Apple M4, 0.604 ms on the CI EPYC) is two loops inside the vendored `mailparse`: `data_encoding::decode_base_mut` (~53%) and the whitespace strip inlined into `decode_base64` (~27%), with `memmem` at ~7.5%. Neither loop can be timed in isolation, and every experiment on them costs a `maturin develop --release` plus an interleaved pytest run. On the x86 runners that wall-time number is also confounded by code layout: identical instructions measured 0-96% apart depending on where the linker placed a loop, and CONTRIBUTING.md says outright that the gate "has a false-positive mode its noise floor cannot see".

There is also no allocation count and no peak-live-bytes number for any mode. Allocation count and peak bytes are deterministic and layout-independent, so they are the natural regression guard for exactly the work that remains (the decode path allocates a strip scratch buffer, a decoded buffer and a `PyBytes` copy per base64 part -- roughly 3.1x the part size in allocation traffic), and for the memory claims lazy and metadata modes make (`raw: part.raw_bytes.to_vec()` is "the one copy lazy mode makes"; "Metadata mode keeps nothing"; "the second read allocates nothing"). Today those claims live in comments and the CHANGELOG; the one memory test in the suite measures `ru_maxrss` on the oversized-rejection path only, on Linux only.

The fuzz harness already shows how to reach the core without Python symbols: it is a standalone crate that includes `src/mail_parser.rs` by path. A bench crate can reuse that pattern with zero new dependencies.

## Evidence

- `Cargo.toml:32-35` -- `[lib] name = "fast_mail_parser" path = "src/fast_mail_parser.rs" crate-type = ["cdylib"]`. The root crate cannot host `cargo bench`/`cargo test` targets that reach the core without Python symbols, and cannot be a Rust dependency.
- `Cargo.toml:25-30` -- `[profile.release] opt-level = 3 / debug = false / strip = "debuginfo" / lto = true / codegen-units = 1`. Function names survive (the symbol table is kept, which is how the maintainer's sample attributed 53% to `decode_base_mut`), but inline frames and line-level attribution do not: a sample inside the strip loop lands on `decode_base64` or `get_body_raw`, not on a line of `bytescan.rs`.
- `fuzz/Cargo.toml:1-9` -- "Deliberately a standalone crate, not a workspace member ... the target includes `src/mail_parser.rs` by path. That works because the core genuinely has no PyO3 dependency -- it uses only `charset`, `mailparse` and `std`". `fuzz/fuzz_targets/parse_email.rs:22-23` -- `#[path = "../../src/mail_parser.rs"] mod mail_parser;`. `fuzz/Cargo.toml:47-48` -- `[patch.crates-io] mailparse = { path = "../vendor/mailparse" }`.
- `src/mail_parser.rs:308, 577, 741, 752, 963, 1055` -- `pub(crate) fn parse_email`, `parse_email_metadata`, `decode_part`, `parse_email_lazy`, `parse_email_tree`, `parse_tree_deferred(payload, defer)`. Inside a crate that includes the module by path, `pub(crate)` is exactly the visibility a bench needs.
- `vendor/mailparse/src/bytescan.rs:65-66` -- `pub(crate) fn strip_ascii_whitespace(body: &[u8]) -> Vec<u8> { let mut cleaned = Vec::with_capacity(body.len());` -- a full-part-size scratch allocation per base64 part. `vendor/mailparse/src/body.rs:139-142` -- `fn decode_base64(body) { let cleaned = crate::bytescan::strip_ascii_whitespace(body); Ok(data_encoding::BASE64_MIME_PERMISSIVE.decode(&cleaned)?) }` -- a second, decoded-size allocation. `src/fast_mail_parser.rs:376, 593, 646, 1052` -- `PyBytes::new(py, decoded.as_slice())` -- a third copy at the boundary.
- `vendor/mailparse/src/bytescan.rs:1-49` -- the module has no `use crate::...` line; its only dependencies are `core::mem::size_of` and its own private helpers. It can be included by path from another crate without a feature flag.
- `src/mail_parser.rs:838` -- `raw: part.raw_bytes.to_vec(),` with the comment "The one copy this mode makes"; `src/mail_parser.rs:1112-1114` -- "Metadata mode keeps nothing." / `raw: defer.then(|| part.raw_bytes.to_vec()),`. `CHANGELOG.md:127` -- "lazy mode retains a copy of every leaf ... metadata mode retains none. On a tree that difference is the memory"; `CHANGELOG.md:183` -- "the second read allocates nothing." None of these has an allocation test.
- `tests/test_payload_memory.py:27-30, 62-76` -- the only memory test: `skipif(not sys.platform.startswith("linux"))`, and it measures `ru_maxrss` growth while rejecting a 100 MiB oversized payload. `tests/test_lazy_mode.py:220-237` asserts `is_decoded` flags, not bytes.
- `CONTRIBUTING.md:355-363` -- "The gate has a false-positive mode its noise floor cannot see ... every treatment benchmark moves together, tightly and repeatably. ... the base64 decode proper (`data_encoding`) and charset conversion are also loops whose placement the linker decides, and nobody has measured how sensitive they are."
- `pyproject.toml:46-52` -- `include = [ { path = "src/*", format = "sdist" }, { path = "vendor/mailparse/**/*", format = "sdist" } ]` and `exclude = ["fuzz/**"]`. A new top-level crate must be added to `exclude` or it ships in the sdist.
- `.github/workflows/test.yml:62-64` -- the vendored-tests step uses `--target-dir "$RUNNER_TEMP/vendor-target"` because a `target/` under `vendor/mailparse` would be swept into the sdist by the include glob. `.github/workflows/test.yml:169-170` -- `cargo deny check` runs on the root lockfile only; `fuzz/` has no committed `Cargo.lock` (`ls -a fuzz` shows `Cargo.toml` and `fuzz_targets` only), so a second crate's dependency tree is already outside the supply-chain check. A std-only bench crate does not widen that.
- Repo-wide grep for `global_allocator`, `GlobalAlloc`, `criterion`, `divan`, `[[bench]]`, `samply`, `line-tables`, `profil` over `*.rs *.toml *.md *.py *.yml *.sh Makefile`: no hits except `[profile.dev]`/`[profile.release]` and the fuzz crate's `[profile.release] debug = true` (`fuzz/Cargo.toml:41-42`), which is itself the precedent for a debuggable release build.

## Proposal

Add `bench/`, a standalone crate mirroring `fuzz/` (not a workspace member, `publish = false`, `[patch.crates-io] mailparse = { path = "../vendor/mailparse" }`, `charset = "0.1.3"` and `mailparse = "0.16.1"` matching the root). **New dependencies: none.** Timing uses `std::time::Instant`; allocation counting uses a `std::alloc::GlobalAlloc` wrapper over `System`. `criterion`/`divan` are deliberately not used: they would be the only crates in the repository outside `cargo deny`'s reach, and the numbers this crate exists to produce (medians, allocation counts) do not need them.

Three binaries, all including the core by path exactly as the fuzz targets do:

1. `bench/src/main.rs` -- a timing loop. `cargo run --release -- <fixture> <mode> [--rounds N]` parses the fixture N times (warm-up, then N rounds), prints min/median/p90 in microseconds. Modes: `full` (`parse_email`), `metadata` (`parse_email_metadata`), `lazy` (`parse_email_lazy`), `tree` (`parse_email_tree`), `tree-lazy`/`tree-metadata` (`parse_tree_deferred(payload, true|false)`), `decode-part` (`decode_part` on the raw bytes of the first attachment, extracted once via `mailparse::parse_mail`), and two isolated loops: `strip` (`bytescan::strip_ascii_whitespace` on the first attachment's encoded body, via `#[path = "../vendor/mailparse/src/bytescan.rs"] mod bytescan;`) and `b64` (`data_encoding::BASE64_MIME_PERMISSIVE.decode` on the already-stripped bytes -- `data-encoding = "2.6.0"` is already in the root lockfile at that exact version; pin the same). The `strip` + `b64` pair is the ~80% of a full parse, timed separately for the first time.
2. `bench/tests/allocs.rs` -- an integration test with its own `#[global_allocator]` counting `alloc` calls, bytes requested, and peak live bytes (an `AtomicUsize` triple; `realloc` counts as one call and adjusts live bytes by the delta). For every `(fixture, mode)` over `tests/data/{attachment,valid,invalid,large}_message.eml` it records `(allocs, bytes, peak)` for the parse alone (reset counters after reading the file, read after the result is built and before it is dropped) and compares against a committed expectation table with **exact equality**, printing the actual triple on mismatch. Exact, not bounded, so any change in allocation behaviour is a visible diff to review; a legitimate change (toolchain bump moving `Vec` growth, a deliberate copy removed) updates the table in the same PR with the before/after in the body. Plus three relational assertions that turn the prose claims into tests, independent of the exact table: `peak(metadata) < peak(full)` and `allocs(metadata) < allocs(full)` on every fixture; `peak(lazy_untouched) < peak(full)` on `large_message.eml` (the fixture with four base64 attachments); `peak(tree_metadata) < peak(tree_lazy)` on `large_message.eml`. Expected but unmeasured: metadata mode's peak on the 767 KiB fixture should be well under a quarter of the input, since it copies no part bodies -- the first run of this test is the measurement that confirms or corrects that.
3. Wire it into the lint job: `cargo fmt --check`, `cargo clippy --all-targets -D warnings`, and `cargo test --locked` inside `bench/`, each with `--target-dir "$RUNNER_TEMP/bench-target"` for the same reason as `test.yml:62-64`. Commit `bench/Cargo.lock` so the bench resolves the *same* `data-encoding`, `encoding_rs`, `memchr`, `quoted_printable` versions the wheel ships -- unlike the fuzz crate, this crate's numbers are only meaningful if its dependency tree matches the root's; add a lint-step check that every `name/version` pair in `bench/Cargo.lock` that also appears in the root `Cargo.lock` has the same version (a 10-line Python or `awk` comparison is enough).

Separately, make the shipped codegen profileable at line level, without touching `[profile.release]`:

```toml
# Cargo.toml -- same codegen as release, plus line tables for samply/perf.
[profile.profiling]
inherits = "release"
debug = "line-tables-only"
strip = "none"
```

and document `maturin develop --profile profiling` in CONTRIBUTING.md's Performance section, with the equivalent that needs no manifest change at all: `CARGO_PROFILE_RELEASE_DEBUG=line-tables-only CARGO_PROFILE_RELEASE_STRIP=none maturin develop --release`. Neither changes what `pip install` or `publish.yml` builds. Be explicit in the doc that a `profiling` build is a different binary from the `release` one (line tables change section sizes and therefore layout), so it is for attributing time, not for measuring it -- the interleaved pytest gate remains the timing oracle for the wheel.

Behaviour that must stay identical: everything. This change adds no code to `src/`, `vendor/`, or `fast_mail_parser/`; the wheel and sdist are byte-for-byte what they were, modulo the `pyproject.toml` `exclude` line.

1. Create `bench/Cargo.toml`: `[package] name = "fast-mail-parser-bench" version = "0.0.0" edition = "2021" publish = false`; `[dependencies] charset = "0.1.3"`, `mailparse = "0.16.1"`, `data-encoding = "2.6.0"`; `[[bin]] name = "fmp-bench" path = "src/main.rs" test = false bench = false`; `[profile.release] debug = "line-tables-only"` (so `cargo run --release` output is itself profileable); `[patch.crates-io] mailparse = { path = "../vendor/mailparse" }`. Copy the header comment from `fuzz/Cargo.toml:1-12` and adapt it: the two dependency versions must match the root or the bench measures something other than what ships.
2. `bench/src/main.rs`: `#[path = "../../src/mail_parser.rs"] mod mail_parser;` and `#[path = "../../vendor/mailparse/src/bytescan.rs"] mod bytescan;`. `bytescan.rs` starts with an inner `#![allow(clippy::chunks_exact_to_as_chunks)]`, which is legal at the top of a path-included module file (the fuzz targets already include `mail_parser.rs`, which opens with `//!` doc comments). Expect `dead_code` warnings for core functions the bench does not call; silence them with `#[allow(dead_code)]` on the `mod` items, as a standalone consumer of a larger module must. Read fixtures relative to `CARGO_MANIFEST_DIR/../tests/data/` so `cargo run` works from any cwd.
3. To get "the first attachment's encoded body" for the `strip`/`b64`/`decode-part` modes: `mailparse::parse_mail(payload)?` then walk `subparts` to the first part whose `Content-Disposition` is `attachment` and take `part.raw_bytes` (for `decode-part`) and the body slice after the header/body separator (for `strip`); simplest is to reuse `mail_parser::decode_part`'s own approach -- read `src/mail_parser.rs:741-750` first. Do the extraction once, outside the timed loop.
4. `bench/tests/allocs.rs`: define `struct Counting; unsafe impl GlobalAlloc for Counting` delegating to `System`, with `ALLOCS`, `BYTES`, `LIVE`, `PEAK: AtomicUsize` (peak updated with a `fetch_max`), and `#[global_allocator] static A: Counting = Counting;`. Reset all four before the parse under test, read after. Include `mail_parser.rs` by the same `#[path]`. Run tests single-threaded (`#[test]` functions that read the counters must not interleave) -- either one `#[test]` that iterates the whole table, or `--test-threads=1` in the CI command; the former is more robust. The expectation table is a `const` array of `(fixture, mode, allocs, bytes, peak)`; on mismatch, print the full actual table in the format of the source so a maintainer can paste it.
5. `pyproject.toml:52`: `exclude = ["fuzz/**", "bench/**"]`. Then confirm with `maturin sdist` (into a scratch dir) + `tar tzf` that no `bench/` path appears; `tests/`/CI already have `.github/scripts/sdist_smoke.py` for a source install, so also run that flow once locally.
6. `.gitignore`: add `/bench/target` (mirror the `/vendor/mailparse/target` entry at the bottom of the file). Do not add `bench/Cargo.lock` to `.gitignore` -- it is committed on purpose (see Proposal).
7. `.github/workflows/test.yml`, `lint` job: after the `vendored mailparse tests` step (line 62-64), add `bench crate` steps: `cargo fmt --manifest-path bench/Cargo.toml --all -- --check`, `cargo clippy --manifest-path bench/Cargo.toml --all-targets --target-dir "$RUNNER_TEMP/bench-target" -- -D warnings -W clippy::cast_possible_truncation`, `cargo test --manifest-path bench/Cargo.toml --locked --release --target-dir "$RUNNER_TEMP/bench-target"` (release, because the allocation counts must be those of optimised code; debug builds allocate differently in places like `Vec::collect`), and the lockfile-agreement check. Keep `if: always()` and `continue-on-error` conventions of the surrounding steps.
8. `Cargo.toml`: add `[profile.profiling]` as sketched, after `[profile.release]`, with a comment saying it changes layout and is for attribution only. Confirm `maturin develop --profile profiling` is accepted by the maturin version in `pyproject.toml` (`maturin>=1.0,<2.0` -- `--profile` is a pass-through cargo option); if it is not, document only the `CARGO_PROFILE_RELEASE_*` env form.
9. `CONTRIBUTING.md`: in "Performance" (starts line 283), after the paragraph ending at line 363, add a short "Measuring below the FFI" subsection: the `cargo run --release -- large_message.eml full` command, what `strip`/`b64` isolate, the allocation-count test and how to update its table, and the profiling build. Add a one-line mention under "Fuzzing"-style tooling in the Makefile if a `make bench-rust` target is added (optional).
10. `CHANGELOG.md` under `[Unreleased]` -> `### Added`: one entry describing the crate, that it adds no dependencies, and that the wheel/sdist are unchanged.
11. Do **not** touch `vendor/mailparse/` -- the by-path include of `bytescan.rs` is what makes `PATCH.md:29-30` ("`diff -r` against the registry copy shows exactly ...") and the hand-merge list at `PATCH.md:75-87` stay true. Do not touch `fast_mail_parser/__init__.pyi` or `tests/test_contract.py` -- no Python API moves.
12. Traps: never run cargo inside `vendor/mailparse` without `--target-dir` elsewhere (the sdist glob at `pyproject.toml:48` sweeps `target/` in); `bench/` must be in `exclude` before the first `maturin sdist`; the lint job's `dtolnay/rust-toolchain` step installs components for the pinned 1.98.0 -- the bench steps run under that pin too, which is what you want (same compiler as the wheel).

## Harness -- how to prove it

This change adds no code to the extension, so the existing gate is a no-op check, and the new numbers are the deliverable.

New (all from the repo root, toolchain 1.98.0 per `rust-toolchain.toml`):

```bash
# Timing below the FFI. Report median us per mode in the PR body, with the CPU.
cargo run --manifest-path bench/Cargo.toml --release --target-dir /tmp/bench-target -- \
    large_message.eml full metadata lazy tree decode-part strip b64 --rounds 200
# Allocation counts and peak bytes; first run generates the table, later runs assert it.
cargo test --manifest-path bench/Cargo.toml --locked --release --target-dir /tmp/bench-target -- --nocapture
```

Record once, in the PR body: (a) the Rust-level `full` median on `large_message.eml` next to `test__fast_mail_parser___parse_message` from `pytest tests/benchmark -k parse_message --benchmark-min-rounds=25` in the same session -- the difference is the FFI + Python object construction share, never measured before; (b) the `strip` and `b64` medians and their sum as a fraction of `full`, which should reproduce the maintainer's sampled ~27% + ~53% split (if it does not, say so -- that is a finding, not a failure); (c) the full `(fixture, mode) -> (allocs, bytes, peak)` table. These are M4 numbers if produced on the maintainer's laptop; the M4 does not show the x86 layout effect, so for the timing binary state that explicitly. Allocation counts and peak bytes do not depend on layout or CPU and need no x86 confirmation.

Unchanged-extension check (belt and braces, because `pyproject.toml` changed): build both sides and run the interleaved A/B per the existing pattern -- `maturin develop --release` in a venv for master and for the branch, `pytest -q tests/benchmark --benchmark-min-rounds=25 --benchmark-json=<side>-<round>.json` alternating A/B for 3 rounds, then `python .github/scripts/ab_median.py --a-label master --b-label bench --a a1.json a2.json a3.json --b b1.json b2.json b3.json`. Every benchmark in `tests/benchmark/test_read_message.py` must be within the 7% threshold and none is expected to move; controls (`test__mail_parser___*`, `test__mailparser_lib___*`, `test__stdlib_email___*`) set the noise floor. The CI gate (`test.yml` `benchmark` job) confirms on x86.

Full lint/test surface, all must be clean:

```bash
cargo fmt --all -- --check
ruff check .
cargo fmt --manifest-path bench/Cargo.toml --all -- --check
cargo clippy --manifest-path bench/Cargo.toml --all-targets --target-dir /tmp/bench-target -- -D warnings -W clippy::cast_possible_truncation
cargo test  --manifest-path bench/Cargo.toml --locked --release --target-dir /tmp/bench-target
```

`cargo test --manifest-path vendor/mailparse/Cargo.toml --target-dir /tmp/vendor-target` is unaffected (vendor/ is not touched) but costs nothing to run. Plus an sdist content check: `maturin sdist -o /tmp/sdist && tar tzf /tmp/sdist/*.tar.gz | grep -c '^[^/]*/bench/'` must print `0`.

## Acceptance criteria

- [ ] `bench/` exists as a standalone crate with `publish = false`, includes `src/mail_parser.rs` and `vendor/mailparse/src/bytescan.rs` by `#[path]`, patches `mailparse` to `../vendor/mailparse`, and declares no dependency other than `charset 0.1.3`, `mailparse 0.16.1`, `data-encoding 2.6.0` (verified by `cargo tree --manifest-path bench/Cargo.toml -e normal --depth 1`).
- [ ] `bench/Cargo.lock` is committed and every package it shares with the root `Cargo.lock` resolves to the same version; a lint-job step fails if they diverge.
- [ ] `cargo run --release -- large_message.eml <mode>` works for `full`, `metadata`, `lazy`, `tree`, `tree-lazy`, `tree-metadata`, `decode-part`, `strip`, `b64` and prints min/median/p90 in microseconds.
- [ ] `cargo test --release` in `bench/` asserts exact `(allocs, bytes, peak)` for every `(fixture, mode)` over the four `tests/data/*.eml` fixtures against a committed table, and prints the actual table on mismatch.
- [ ] The relational assertions hold and are tested: `peak(metadata) < peak(full)` and `allocs(metadata) < allocs(full)` on all fixtures; `peak(lazy) < peak(full)` and `peak(tree_metadata) < peak(tree_lazy)` on `large_message.eml`.
- [ ] `vendor/mailparse/` is untouched (`git diff --stat master -- vendor/` is empty); `PATCH.md` needs no edit.
- [ ] `pyproject.toml` `exclude` contains `bench/**`; `maturin sdist` output contains no `bench/` path; `.gitignore` covers `/bench/target`.
- [ ] `.github/workflows/test.yml` lint job runs fmt, clippy and `cargo test --locked --release` for `bench/` with `--target-dir "$RUNNER_TEMP/bench-target"`.
- [ ] `Cargo.toml` has `[profile.profiling] inherits = "release" debug = "line-tables-only" strip = "none"`; `[profile.release]` is byte-identical to master; CONTRIBUTING.md documents both `maturin develop --profile profiling` and the `CARGO_PROFILE_RELEASE_*` env form, and says the profiling build is for attribution, not timing.
- [ ] PR body contains: Rust-level medians per mode with CPU named; the FFI share (Rust `full` vs pytest `parse_message`); `strip`+`b64` as a fraction of `full`; the allocation table; and an interleaved `ab_median.py` result showing every pytest benchmark within 7% (expected ~0%).
- [ ] `CHANGELOG.md` `[Unreleased]` has an `### Added` entry; the full lint/test surface above is green; the CI benchmark gate is green.

## Out of scope

- Any change to the decode path itself (in-place strip, single-buffer decode, avoiding the `PyBytes` copy) -- this issue builds the instrument; those are separate issues that should cite its numbers.
- Adding a custom global allocator (mimalloc etc.) to the extension. The allocation table this issue produces is the input to that decision: a full parse that is a few dozen allocations makes it moot; a header-heavy path in the thousands makes it worth an A/B.
- Running the Rust timing binary as a CI gate. Its wall time is subject to the same layout effects on x86 and would need the same interleaving; only the allocation test gates.
- Adopting `criterion`/`divan`. Revisit only if the std timing loop proves insufficient, and then commit and audit the lockfile.
- Changing the pytest benchmark gate, `ab_median.py`, or `check_benchmark.py`.

## Related issues

- #228 — isolates the base64 loop this replaces
- #229
- #230
- #236 — a workspace core crate would let the bench depend on it instead of `#[path]`
- #238

kurok:
Closes #225.

Branched from `master` (`47965f8`). Chosen partly because it adds a new crate rather than touching `src/mail_parser.rs` — #253 and #255 are both in that file, and this way nothing conflicts.

## What it is for

The pytest benchmarks are the oracle for what the wheel costs a user, and a poor instrument for asking *where* that cost is: every number carries the FFI crossing, the GIL and the construction of Python objects. `bench/` measures the core itself.

**Isolating the base64 path.** `strip` and `b64-*` time the whitespace strip and the decode separately — together most of a full parse on the large fixture. `b64-simd` against `b64-scalar` is what #228 bought on its own rather than diluted through a whole parse:

| mode | min (M4) |
|---|---|
| `b64-simd` *(what ships)* | **82.6 µs** |
| `b64-scalar` *(what it replaced)* | 136.8 µs |
| `strip` | 55.7 µs |
| `decode-part` | 101.7 µs |
| `full` | 178.7 µs |
| `metadata` | 33.7 µs |

**1.66×** for #228, measured on its own for the first time.

**Counting allocations.** A global allocator records allocs, bytes and peak live bytes per mode, and the modes' documented memory claims become assertions. On the 785 KiB fixture:

| mode | allocs | bytes | peak |
|---|---|---|---|
| full | 389 | 1,401,650 | **1,060,757** |
| metadata | 346 | 37,985 | **11,929** |
| lazy | 387 | 836,013 | 790,993 |
| tree | 426 | 1,393,164 | 1,061,316 |
| tree-deferred | 423 | 823,108 | 798,111 |

Metadata mode peaks **89× lower** than a full parse. That was expected and never measured.

## Three things the measuring taught me

All now encoded rather than assumed — and each was a failing assertion before it was a finding:

**A lazy tree holds *more* than a decoded one on a small attachment-bearing message** — 6,696 against 6,323 bytes. Lazy retains the *encoded* bytes, and base64 is 4/3 of what it decodes to, so the trade only pays once the bodies are large. That ordering is now asserted only where it is actually a promise.

**Metadata mode's peak ties a full parse's** on a fixture whose bodies are a few hundred bytes, because the peak there is the parse itself. Demanding a strict drop everywhere demands something the mode never promised; `<=` plus a hard assertion on the large fixture is the honest pair.

**The counters are process-global**, so the four separate tests I wrote first counted each other's allocations and failed three assertions very convincingly — it looked exactly like the modes breaking their promises. A mutex does not fix that: it serialises the measuring, not the other threads' allocating. One test does.

## The lockfile check earned itself immediately

`bench/Cargo.lock` is committed and pinned to the versions the wheel ships, with `check_bench_lockfile.py` in the lint job. A freshly resolved lockfile took **`data-encoding 2.11.1` against the shipped 2.6.0** (and `encoding_rs 0.8.41` against 0.8.35) — so the b64 comparison above read **1.86×** until it was pinned, comparing against a decoder the library does not use. The number in this PR is the corrected one.

## Also

A `profiling` cargo profile — release codegen plus line tables — so `samply` and `perf` attribute time to a line instead of a stripped symbol. Documented as a **different binary** from the release one, since line tables change section sizes and therefore layout, and this crate is measurably sensitive to that (#240 measured up to 7.5%). For finding where the time goes, never for measuring how much.

## Checks run locally

`cargo test --release` in `bench/` (passes under parallel `cargo test`, which is the point), `cargo fmt --check` and `cargo clippy --all-targets -D warnings` for the bench crate, `check_bench_lockfile.py`, plus the root surface: `pytest tests --ignore=tests/benchmark` (836 passed, 3 skipped), `cargo fmt --check`, `cargo clippy -D warnings`, `ruff check .`.

No new runtime dependency. Nothing under `src/`, `vendor/` or `fast_mail_parser/` changes — the wheel is byte-identical, and `bench/` is excluded from the sdist alongside `fuzz/`.


kurok:
Rebased onto `097394c` (post-#254 and #255).

**CI was already 15/15 green; the problem was that it had stopped being mergeable, and the rebase alone would not have fixed it.**

`.github/workflows/test.yml` conflicted — both sides add a lint step, so both are kept. But #255 moved `src/mail_parser.rs` to `crates/fast_mail_parser_core/src/lib.rs`, and this crate reached for it by path:

```rust
#[path = "../../src/mail_parser.rs"]   // in both main.rs and tests/allocs.rs
```

A path in a string literal is invisible to a merge. The rebase resolved cleanly and left two dead includes behind it — I only found them by building afterwards rather than trusting the resolution.

So the bench crate now **links** `fast_mail_parser_core` instead, exactly as #255 did for the fuzz harness. That is the better shape anyway, and it is the argument #236 was making: a crate can be depended on, and a dependency survives its source file moving.

Two consequences of #255 that followed:

- `bench` had to join `fuzz` and `vendor/mailparse` in the root `[workspace] exclude`, or cargo adopts it as a member and its pinned lockfile is overridden by the workspace one — which would defeat `check_bench_lockfile.py`. The comment there says so.
- Relinking re-resolved the lockfile, so `data-encoding` and `encoding_rs` needed re-pinning to the shipped versions. The check caught it again; 12 shared packages now agree.

Re-verified after the rebase: bench builds, its tests pass, `cargo fmt --check` and `clippy -D warnings` clean for the bench crate and `--workspace`, `check_bench_lockfile.py` green, 854 Python tests pass, and the timing modes still produce numbers (b64-simd 93.0 µs against b64-scalar 148.9 µs on this run).

PR #258, opened 2026-09-17
Lazy modes borrow the pinned payload instead of copying every deferred leaf (#239)

### LINKED ISSUE #239: Lazy modes: borrow the pinned payload through byte ranges instead of copying every deferred leaf
## Problem

`mode="lazy"` exists to defer work, but its parse still performs one O(message) allocation-and-memcpy: every deferred leaf copies mailparse's `raw_bytes` (headers + still-encoded body) into its own `Vec<u8>`. On the 767 KiB fixture that is the PDF and PNG parts, ~760 KiB of fresh heap per parse, and it is the only thing separating `mode="lazy"` (nothing read) from `mode="metadata"` apart from the two small body decodes: 0.080 ms vs 0.051 ms flat, 0.077 ms vs 0.054 ms for the tree, on the CI runner (`Readme.md:466-473`, `:565-567`).

The copy was made for a lifetime reason, not a speed one: `bytes` payloads are borrowed for the parse (#169), and lazy mode (#198) chose to copy so that "nothing borrows from the caller's `bytes` once this returns" (`src/fast_mail_parser.rs:1391-1395`). But the binding *already* holds a refcounted `PyBackedBytes` handle on the caller's immutable `bytes` for the duration of the parse and simply drops it on return. Keeping that handle alive inside the lazy result and storing byte ranges gives the same lifetime guarantee (the `bytes` object cannot be freed or mutated while referenced) with zero copies, and `parse_many(mode="lazy")` shares one handle per payload across all its attachments.

This is also a memory story: the README warns lazy mode costs 1.33x the decoded size in retained encoded bytes *on top of* the payload the caller usually still holds. With borrowing, a lazy result over a `bytes` payload retains nothing the caller was not already paying for.

## Evidence

All citations are from `master` @ `73f56da`.

- `src/mail_parser.rs:697-707` -- `pub(crate) struct LazyAttachment { ... pub(crate) raw: Vec<u8>, }` (line 706).
- `src/mail_parser.rs:829-838` -- the flat lazy loop pushes `raw: part.raw_bytes.to_vec()` with the comment `// The one copy this mode makes, and what makes it a trade rather than a free win.`
- `src/mail_parser.rs:1107-1116` -- the tree does the same per leaf: `raw: defer.then(|| part.raw_bytes.to_vec())`, with `// For the root of a single-part message that is the whole payload`.
- `src/mail_parser.rs:752-766` and `:1055-1068` -- both entry points parse `repaired.as_deref().unwrap_or(payload)`; when `repair_missing_separator` (`:270-306`) fires, `raw_bytes` point into a *local* `Vec<u8>`, not the caller's buffer. That local must be returned to the caller in the new design.
- `src/mail_parser.rs:1092-1101` -- in the tree, a `message/rfc822` node decodes its body into `raw`, optionally repairs it, and recurses: `let parsed = parse_mail(repaired.as_deref().unwrap_or(raw.as_slice()))?; build_node(&parsed, depth + 1, defer)?`. Children of an embedded message therefore borrow from a decoded buffer, **not** from the payload. Ranges into the root payload are wrong for them; they need an owned fallback.
- `src/mail_parser.rs:1444-1462` -- `extract_mail_parts` flattens subparts and pushes the root last, so for a single-part non-body message the root itself becomes a `LazyAttachment` whose `raw_bytes` is the whole message.
- `src/mail_parser.rs:741-744` -- `decode_part(raw: &[u8])` re-parses a slice; it takes `&[u8]` and is unaffected by where the slice lives.
- `vendor/mailparse/src/lib.rs:715-718` -- `pub raw_bytes: &'a [u8]`; `:938-946` sets `raw_bytes: raw_data` for the node being parsed; `:970-973` recurses with `&raw_data[ix_part_start..ix_part_end]`. Every part's `raw_bytes` is a subslice of the top-level buffer handed to `parse_mail`, so `raw_bytes.as_ptr() - payload.as_ptr()` is a valid offset (plain `usize` arithmetic, no `unsafe`).
- `src/fast_mail_parser.rs:1188-1191` -- `enum Payload { Borrowed(PyBackedBytes), Owned(Vec<u8>) }`; `:1203-1207` -- `if let Ok(bytes) = obj.cast::<PyBytes>() { return Ok(Payload::Borrowed(PyBackedBytes::from(bytes.clone()))); }`; `:1389-1400` -- `parse_lazy_inner` builds this handle, parses, then lets `message` drop after the core has copied out of it.
- `src/fast_mail_parser.rs:946-947` -- `PyLazyAttachment { ... raw: Vec<u8>, }`; `:1014-1024` -- `from_lazy` moves `raw: attachment.raw`; `:1047-1054` -- `decode` does `let raw = self.raw.as_slice(); py.detach(|| mail_parser::decode_part(raw))`.
- `src/fast_mail_parser.rs:512-514` -- `PyLazyMimePart { ... raw: Option<Vec<u8>>, }` (doc: "The part as it sits in the message, still encoded"); `:625-663` -- `lazy_node` moves `raw` from `NodeBody::Undecoded`.
- `src/fast_mail_parser.rs:1639-1654` -- `parse_many_lazy` builds `Vec<Payload>`, parses with `parse_many_as(&messages, workers, parse_email_lazy)`, and never uses `messages` again.
- `~/.cargo/registry/src/*/pyo3-0.29.2/src/pybacked.rs:194-197` -- `enum PyBackedBytesStorage { Python(Py<PyBytes>), Rust(Arc<[u8]>) }`; `:203-213` -- `pub fn clone_ref(&self, py: Python<'_>) -> Self` is a refcount bump / `Arc::clone`; `:232-233` -- `unsafe impl Send for PyBackedBytes {} unsafe impl Sync for PyBackedBytes {}`; `:261` -- `impl From<Bound<'_, PyBytes>> for PyBackedBytes`. Holding one per attachment object is a refcount, not a copy, and it is fine inside a `#[pyclass]`.
- `tests/test_lazy_mode.py:203-215` -- `test__attachments_outlive_the_payload` does `del payload` then reads `content`. Its comment (`:204-207`) states the design premise: "a mode that retained a view into the caller's buffer would be a use-after-free waiting for a garbage collection. Lazy mode copies each part's encoded bytes". A refcounted handle satisfies the test's *behaviour* (the buffer cannot be collected while referenced); the comment becomes wrong and must be rewritten.
- `fuzz/fuzz_targets/parse_agreement.rs:156-158` (`NodeBody::Undecoded { raw, .. } => ... decode_part(raw)`) and `:293` (`decode_part(&deferred.raw)`) -- the fuzz crate compiles `src/mail_parser.rs` by path (`fuzz/Cargo.toml`), so it must be updated with the core and the core must stay PyO3-free.
- Gate coverage: `.github/scripts/ab_median.py:36-40` names only `test__mail_parser___`, `test__mailparser_lib___`, `test__stdlib_email___` as controls, and `.github/workflows/test.yml:466-471` passes only `--informational test__threaded___`. **Every other benchmark, including `test__fast_mail_parser___parse_lazy_untouched` (`tests/benchmark/test_read_message.py:283`) and `..._parse_tree_lazy_untouched` (`:355`), is a gate treatment**, and the verdict is the *worst* treatment delta at `AB_THRESHOLD_PCT=7`.
- Fixture: `tests/data/large_message.eml` is 785,774 bytes; its deferred leaves are the `application/pdf` part (line 145) and the `image/png` part (line 3946), i.e. essentially the whole file. Bodies are a small 8bit `text/plain` (line 39) and a ~7 KiB base64 `text/HTML` (line 56).
- Documented memory semantics that change: `Readme.md:483-486` ("The encoded bytes of every attachment are retained until the message is dropped"), `Readme.md:561-564` ("lazy mode keeps a copy of every leaf"), `fast_mail_parser/__init__.pyi:135-137`, `:526-528`, `:685-688`.

## Proposal

**Core (`src/mail_parser.rs`, stays PyO3-free):**

```rust
/// Where a deferred leaf's encoded bytes live.
#[derive(Debug)]
pub(crate) enum Retained {
    /// A sub-range of the buffer the message was parsed from.
    Range(std::ops::Range<usize>),
    /// An owned copy: only for leaves parsed out of a decoded `message/rfc822`
    /// body, whose bytes are not in the caller's buffer.
    Owned(Vec<u8>),
}

pub(crate) struct LazyAttachment { ..., pub(crate) raw: Retained }        // was Vec<u8>
pub(crate) struct LazyMail { ..., pub(crate) repaired: Option<Vec<u8>> } // NEW: the #150 copy, if one was made
pub(crate) enum NodeBody { ..., Undecoded { encoded_size: usize, raw: Option<Retained> } }
pub(crate) struct DeferredTree { pub(crate) root: TreeNode, pub(crate) repaired: Option<Vec<u8>> } // or add the field to the root

/// Offsets of `part` inside `base`. Sound because mailparse's `raw_bytes` is
/// always a subslice of the buffer given to `parse_mail` (lib.rs:970-973).
fn range_in(base: &[u8], part: &[u8]) -> std::ops::Range<usize> {
    let start = part.as_ptr() as usize - base.as_ptr() as usize;
    debug_assert!(start + part.len() <= base.len());
    start..start + part.len()
}

impl Retained {
    pub(crate) fn slice<'a>(&'a self, base: &'a [u8]) -> &'a [u8] {
        match self { Retained::Range(r) => &base[r.clone()], Retained::Owned(v) => v }
    }
}
```

- `lazy_from_payload(payload)` computes `Retained::Range(range_in(payload, part.raw_bytes))` instead of `to_vec()`; `parse_email_lazy` moves the `repaired` local into `LazyMail::repaired` instead of dropping it.
- `build_node` gets a `base: Option<&[u8]>` parameter: `Some(root buffer)` while walking the top-level message, `None` once it recurses into a decoded `message/rfc822` body (`:1092-1101`), where it keeps `Retained::Owned(part.raw_bytes.to_vec())`. `parse_tree_deferred` returns the repaired copy alongside the root, as the flat path does.
- `decode_part` is unchanged.

**Binding (`src/fast_mail_parser.rs`):**

```rust
/// The buffer a lazy result's ranges index into. Shared by refcount across
/// every attachment object of one parse (and across a whole parse_many batch
/// slot). `PyBackedBytes` is `Send + Sync` and `clone_ref` is a refcount bump.
enum SharedPayload { Python(PyBackedBytes), Rust(Arc<[u8]>) }

impl Payload {
    /// The buffer the parse ran over, pinned for the result's lifetime.
    /// `repaired` wins when present: the ranges were computed against it.
    fn into_shared(self, repaired: Option<Vec<u8>>) -> SharedPayload { ... }
}

struct PyLazyAttachment { ..., buffer: SharedPayload, raw: Retained /* or Range + Option<Vec<u8>> */, content: OnceLock<Py<PyBytes>> }
```

- `PyLazyAttachment::decode` becomes `let raw = self.raw.slice(self.buffer.as_ref()); py.detach(|| decode_part(raw))` -- identical output.
- `PyLazyMail::from_lazy(py, mail, buffer: &SharedPayload)` hands each attachment a `clone_ref`/`Arc::clone` of the buffer. `parse_lazy_inner`, `parse_many_lazy` (zip results with the consumed `messages`), and `parse_email_tree_other_mode`/`lazy_node` thread the buffer through the same way.
- `str` payloads: `Payload::Owned(Vec<u8>)` becomes `SharedPayload::Rust(Arc::from(vec))` -- still one copy of the message (as today), but no *second* copy per attachment. If round-1's "Borrow str payloads through PyBackedStr" lands first, str input rides the `Python` variant for free.

**New dependency:** none. `std::sync::Arc`, `std::ops::Range`, and `pyo3::pybacked::PyBackedBytes` (already imported at `src/fast_mail_parser.rs:21`).

**Behaviour that must stay identical:** every value of every attribute on `PyLazyMail`, `PyLazyAttachment`, `PyLazyMimePart` (the frozen attribute sets in `tests/test_contract.py` and `tests/test_lazy_mode.py:425`); `content is content`; `is_decoded` semantics; `DecodeError` raised on access, not cached; `warnings`/`strict` parity with full mode; the #150 separator warning; `MAX_INPUT_BYTES`/`MAX_MIME_DEPTH` caps; attachments outliving `del payload`.

**Behaviour that changes (document it):** a lazy result now keeps the *entire* payload alive (the caller's `bytes` object, or one owned copy of the message for `str`/repaired input) instead of a copy of each deferred part. For attachment-heavy mail this is less memory; for a message with a huge text body and one tiny attachment it is more, if the caller drops the payload and keeps the lazy result.

1. **Core first** (`src/mail_parser.rs`): add `Retained` + `range_in` + `Retained::slice`; change `LazyAttachment.raw`, add `LazyMail.repaired`; change `NodeBody::Undecoded.raw` to `Option<Retained>`; give `build_node` the `base: Option<&[u8]>` argument and return the repaired buffer from `parse_tree_deferred`. Keep every `#[inline(never)]` that is there (`lazy_from_payload`, `build_node`, `repair_missing_separator`) -- `CONTRIBUTING.md:283-330` explains why cold code is kept out of line. Do not touch `Mail::new`/`from_payload` or `parse_email_metadata`: the flat full path must not change shape.
2. **Fuzz target** (`fuzz/fuzz_targets/parse_agreement.rs:156-158`, `:293`): resolve through `Retained::slice(&lazy.repaired.as_deref().unwrap_or(data))` so the target still asserts "the retained slice is the part" on arbitrary input. `fuzz/` compiles the core by path; `cargo check` it (`cargo +nightly fuzz check` or `cargo check --manifest-path fuzz/Cargo.toml --target-dir /tmp/fuzz-target`) even though CI only runs it weekly.
3. **Binding** (`src/fast_mail_parser.rs`): add `SharedPayload`, `Payload::into_shared`; update `PyLazyAttachment` (`:946-947`, `:1014-1024`, `:1047-1054`), `PyLazyMimePart` (`:512-514`, `:551-558`, `:568-571`, `:589-595`), `lazy_node` (`:625-663`), `PyLazyMail::from_lazy` (`:1137-1161`), `parse_lazy_inner` (`:1389-1400`, rewrite its comment), `parse_many_lazy` (`:1639-1680`, consume `messages` with `into_iter().zip(parsed)`), `parse_email_tree_other_mode` (`:1722-1745`). `is_decoded` on `PyLazyMimePart` (`:568-571`) keys off `raw.is_none()` -- keep that. `#[pyclass]` requires `Send`; `PyBackedBytes` and `Arc<[u8]>` are. No `__traverse__` is needed: `bytes` cannot form reference cycles.
4. **Do not** compute offsets in the binding or hold a raw `*const u8` across `py.detach`; the range is computed in the core against the exact buffer the core parsed, and the binding only indexes.
5. **Tests**: `tests/test_lazy_mode.py` (all of it; rewrite the comment at `:204-207` -- the behaviour it asserts still holds, the reason is now the refcount), `tests/test_tree_lazy*.py` / whatever covers `parse_email_tree(mode="lazy")` including the embedded `message/rfc822` fixtures in `tests/data/rfc/`, `tests/test_metadata_mode.py`, `tests/test_parse_many.py`, `tests/test_stdlib_parity.py` (its `DIVERGENCES` list must not grow), `tests/test_contract.py` (no new Python-visible attribute, so nothing to add).
6. **New tests** (add to `tests/test_lazy_mode.py` / `tests/test_payload_memory.py`):
   - `test__a_lazy_parse_does_not_copy_the_attachments`: Linux-only `ru_maxrss` subprocess probe in the style of `tests/test_payload_memory.py:33-58`: build a ~100 MiB single-attachment message as `bytes`, parse with `mode="lazy"`, assert peak growth < 20 MiB (today it is ~100 MiB).
   - `test__the_payload_object_is_kept_alive_by_a_lazy_result`: `sys.getrefcount(payload)` rises by one after `parse_email(payload, mode="lazy")` and falls back after `del mail` -- makes the new memory contract observable.
   - `test__nested_message_leaves_still_decode_after_del_payload` for the tree: an `.eml` with a `message/rfc822` part whose inner message has a base64 attachment; parse lazily, `del payload`, decode the inner leaf, compare with full mode.
   - `test__repaired_messages_decode_in_lazy_mode`: `tests/data/invalid_message.eml` (header block never terminated, #150) round-trips through both lazy modes and equals full mode -- this is the `repaired` buffer path.
7. **Docs**: `Readme.md:483-486` and `:561-564` (say: "a lazy result keeps the payload alive; over `bytes` it retains nothing the caller was not already holding; a `str` payload or a message that needed the #150 repair is held as one owned copy"); `fast_mail_parser/__init__.pyi:135-137`, `:526-528`, `:685-688`; the `parse_lazy_inner` comment; `tests/benchmark/test_read_message.py:307-311` and `:356-358` comments ("It retains a copy of every leaf"). `CHANGELOG.md` under `## [Unreleased]` / `### Changed`, with the local before/after numbers.
8. **Traps**: no change under `vendor/` (so `PATCH.md` and the sdist include glob at `pyproject.toml:46-48` are untouched -- verify with `git status vendor/`); abi3: `PyBackedBytes::from(Bound<PyBytes>)` is already used under `abi3-py311`, so no new limited-API surface; `clippy -W clippy::cast_possible_truncation` -- the pointer subtraction is `usize`, no cast; never run cargo inside `vendor/mailparse` without `--target-dir`.

## Harness -- how to prove it

Two venvs, two wheels, same interpreter:

```bash
cd /path/to/fast_mail_parser
git worktree add ../fmp-base master
for side in base head; do
  src=$([ $side = base ] && echo ../fmp-base || echo .)
  uv venv .venv-$side && uv pip install --python .venv-$side/bin/python maturin pytest pytest-benchmark mail-parser
  (cd $src && ../fast_mail_parser/.venv-$side/bin/maturin build --release -o ../fast_mail_parser/dist-$side)
  uv pip install --python .venv-$side/bin/python dist-$side/*.whl
done
for round in 1 2 3 4 5; do for side in head base; do
  .venv-$side/bin/pytest -q tests/benchmark --benchmark-min-rounds=25 --benchmark-json=$side-$round.json
done; done
```

- **Expected to improve:** `test__fast_mail_parser___parse_lazy_untouched` and `test__fast_mail_parser___parse_tree_lazy_untouched` -- expected -15% to -30%, bounded above by the gap to `..._parse_metadata` / `..._parse_tree_metadata` in the same run (0.029 ms / 0.023 ms on the CI EPYC). The gap that remains should be roughly the two body decodes. `test__fast_mail_parser___parse_many_metadata` is unaffected; a `parse_many(mode="lazy")` benchmark does not exist yet -- if round-1's benchmark issues have not added one, add `test__fast_mail_parser___parse_many_lazy_untouched` (8 x fixture, `threads=1`, guarded with the same `try/except (TypeError, ValueError)` skip as `:373-389`) so the shared-handle path is measured.
- **Must not regress (> 7%):** everything else, in particular `test__fast_mail_parser___parse_message` (the gate's headline and `check_benchmark.py`'s absolute floor), `..._parse_message_strict`, `..._parse_tree`, `..._parse_metadata`, `..._parse_lazy_all_attachments` (decode cost is unchanged; only the copy is gone, so it should improve slightly or hold).
- **Memory:** run the new `ru_maxrss` probe on Linux (a container is fine); report before/after growth in the PR.
- **Full local CI surface:** `pytest tests --ignore=tests/benchmark`, `cargo fmt --all -- --check`, `cargo clippy --all-targets -- -D warnings -W clippy::cast_possible_truncation`, `mypy --strict fast_mail_parser/`, `ruff check .`, `cargo test --manifest-path vendor/mailparse/Cargo.toml --target-dir /tmp/vendor-target` (unchanged vendor, but run it), plus a `cargo check` of `fuzz/` with its own `--target-dir`.
- Local before/after `ab_median.py` tables go in the PR body with the CPU line. **The M4 is not a proxy for x86**: every `test__fast_mail_parser___*` benchmark is a gate treatment and the flat parse has swung up to 96% from layout alone on Zen (#204). If the PR gate reports a regression on a benchmark this change does not touch, re-run it and/or run `toolchain-ab.yml` before drawing a conclusion; do not chase it by moving code in the hot path.

## Acceptance criteria

- [ ] `src/mail_parser.rs` no longer contains `part.raw_bytes.to_vec()` on the top-level lazy paths (flat loop at ~`:836`, tree leaf at ~`:1114`); the only remaining copy of a deferred leaf is for leaves under a decoded `message/rfc822` body, and it is commented as such.
- [ ] `LazyMail` and the deferred tree return the #150 repaired buffer to the caller instead of dropping it; `tests/data/invalid_message.eml` round-trips through `parse_email(mode="lazy")` and `parse_email_tree(mode="lazy")` equal to full mode.
- [ ] `PyLazyAttachment` and `PyLazyMimePart` index a shared, refcounted buffer; `parse_many(mode="lazy")` creates exactly one shared handle per input slot.
- [ ] `tests/test_lazy_mode.py::test__attachments_outlive_the_payload` passes unchanged in behaviour, with its comment rewritten to the refcount rationale.
- [ ] New tests pass: `ru_maxrss` lazy-parse probe (< 20 MiB growth for a 100 MiB single-attachment `bytes` payload), refcount-observability test, nested `message/rfc822` leaf after `del payload`, repaired-message lazy round-trip.
- [ ] `fuzz/fuzz_targets/parse_agreement.rs` compiles against the new core API and still asserts deferred-decode == full-decode for both flat and tree lazy modes.
- [ ] Interleaved local A/B (>= 3 rounds, `--informational test__threaded___`): `..._parse_lazy_untouched` and `..._parse_tree_lazy_untouched` improve; no treatment benchmark regresses > 7%; the table with CPU line is in the PR body.
- [ ] `Readme.md`, `fast_mail_parser/__init__.pyi` (three docstrings), benchmark comments and `CHANGELOG.md` describe the new memory contract (payload pinned; `str`/repaired input held as one owned copy; nested-message leaves copied).
- [ ] `git status vendor/` is clean; `tests/test_contract.py` unchanged; `tests/test_stdlib_parity.py` `DIVERGENCES` unchanged; full lint/test surface green.

## Out of scope

- Borrowing `str` payloads (round-1: "Borrow str payloads through PyBackedStr instead of copying them") -- this issue only needs `Payload::Owned` to become an `Arc`; if that lands first, the `str` path inherits the `Python` variant.
- Full-mode attachment copies on `.attachments`/`.content` reads (round-1: "Full-mode attachments deep-copied on every mail.attachments read").
- Letting nested `message/rfc822` children borrow from the parent's published `content` `PyBytes` instead of copying -- possible via `PyBackedBytes`, but the inner message may have been re-parsed from a repaired copy (`src/mail_parser.rs:1097-1098`), so the offsets would not match; leave as an owned copy.
- Any change to `decode_part`, the vendored mailparse, or full/metadata modes.
- A `parse_many` thread-scaling benchmark (round-1: "Benchmarks: parse_many thread-scaling curve").

## Related issues

- #226
- #227
- #237
- #224

kurok:
Closes #239.

Deferring a decode used to copy the part's encoded bytes out of the message, which is the opposite of what deferring is for. A deferred part now keeps its **offsets** in the buffer it was parsed from, and the result keeps that buffer alive.

## What it costs and saves

`ru_maxrss` of a subprocess parsing a 96 MiB single-attachment message in `mode="lazy"` and reading nothing:

| | peak growth |
| --- | --- |
| master | 192.3 MiB — the payload, plus a copy of the attachment |
| this PR | 96.2 MiB — the payload |

Counted in the core (`bench/tests/allocs.rs`, peak live bytes), before → after:

| fixture / mode | master | this PR |
| --- | --- | --- |
| `large_message.eml` lazy | 790,993 | **23,398** |
| `large_message.eml` tree-lazy | 798,111 | **14,332** |
| `valid_message.eml` tree-lazy | 113,597 | **15,356** |
| `attachment_message.eml` tree-lazy | 6,696 | **6,295** |

A lazy tree now allocates exactly what a metadata tree allocates, because a range is not a copy. The old table's note that a lazy tree could hold *more* than a fully decoded one — 6,696 against 6,323 bytes, base64 being 4/3 of what it decodes to — no longer describes anything, and the assertion that replaces it is the one that was false before: a lazy tree's peak is under a quarter of the input.

## The contract this changes

**A deferred part pins the payload.** A `PyLazyAttachment` or a `PyLazyMimePart` leaf keeps the message it was parsed from alive for as long as it is reachable, so keeping one attachment out of a mailbox keeps that message rather than just the part. Values are unchanged in every mode. The pin is per message even in `parse_many` — one slot never holds another's payload — and a caller who wants the bytes without the message reads `content`, which is a decoded copy, and drops the attachment. `Readme.md`, the three `__init__.pyi` docstrings and the `CHANGELOG` all say so.

## Three things the ranges have to get right

Each has its own guard, because each fails silently.

- **A repaired message is parsed from a rebuilt copy.** The #150 separator repair means the offsets index the rebuild, not the caller's payload — so `LazyMail` and the new `DeferredTree` carry that buffer instead of dropping it. A core test asserts the rebuild is one byte longer *and* that resolving against it gives the right bytes, since resolving against the payload would also "work" and return something one byte off.
- **A leaf below a `message/rfc822` node has nothing to point into.** Its bytes came from decoding that body, which is dropped when the walk leaves the subtree, so it keeps a copy. `Retain` has three states rather than two for exactly this reason: "keep nothing" and "keep a copy because there is no buffer" are different answers, and my first version conflated them — which dropped those leaves' bytes entirely and turned lazy mode into metadata mode for that subtree. A core test caught it; the Python suite did not, so it has one now too.
- **An offset that is off by one still decodes, to something else.** `Retained::of` is bounds-checked rather than asserted, so it can never index out of range — but its fallback is a silent copy, which would degrade to exactly the behaviour this issue removes. The PR-time fuzz job therefore runs with `-a` (debug assertions), where that fallback is a panic. `deep-fuzz` stays on the release build, where executions per second matter more.

## Measurements

Local interleaved A/B, Apple M4, 3 rounds per side, `ab_median.py`, **run twice**:

| Benchmark | master | this PR | |
| --- | --- | --- | --- |
| `parse_lazy_untouched` | 0.041 ms | **0.030 ms** | −26% |
| `parse_tree_lazy_untouched` | 0.040 ms | **0.030 ms** | −25% |
| `parse_lazy_all_attachments` | 0.173 ms | 0.164 ms | −5% |

Nothing else moved past the control noise floor (1.2% and 2.7% in the two runs); the worst non-target value was −1.9%. Both runs reproduced the two headline numbers within 0.7 points of each other.

**Verification**: 864 Python tests, 10 core Rust tests (which run in debug, where `Retained::of`'s assertion is live), `bench/tests/allocs.rs`, `cargo clippy --all-targets -D warnings` on all three crates, `cargo fmt`, `mypy --strict`, `ruff`, and the vendored-mailparse and bench-lockfile checks. 527K fuzz executions of `parse_agreement` with debug assertions on, plus the other three targets, no findings.

**Note for anyone building locally on macOS 26/27**: `strip = "debuginfo"` in `[profile.release]` currently produces a `.so` CPython refuses to load (`mis-aligned LINKEDIT string pool`). It reproduces on master, so it is not from this change; `CARGO_PROFILE_RELEASE_STRIP=none` works around it, and both sides of the A/B above were built that way.


kurok:
### The benchmark gate failed once and passed on re-run — both numbers, for the record

**Run 1** (failed, worst delta +16.3%): the regressions were `parse_tree_metadata` +16.3%, `parse_many_metadata` +12.7%, `parse_metadata` and `parse_metadata_str` +11.2% each, `headers_first_read` +7.1%. Noise floor from the controls: **3.9%**.

Four of those five run code this PR cannot reach. `parse_metadata` and `parse_metadata_str` go through `parse_email_metadata` → `metadata_from_payload`, which this diff does not touch and which never calls `build_node`; `headers_first_read` is a full parse; `parse_many_metadata` is `parse_many_as(..., parse_email_metadata)`. The only changed thing on any metadata path is `build_node`'s third parameter, which the tree walk takes and the flat metadata path does not have.

**Run 2** (passed, worst delta +1.8%, noise floor **0.8%**), per-round minima:

| Benchmark | base | this PR | |
| --- | --- | --- | --- |
| `parse_lazy_untouched` | 0.094 / 0.095 / 0.093 | 0.055 / 0.055 / 0.054 | **−42%** |
| `parse_tree_lazy_untouched` | 0.091 / 0.092 / 0.091 | 0.053 / 0.054 / 0.053 | **−42%** |
| `parse_metadata` | 0.050 / 0.051 / 0.050 | 0.050 / 0.051 / 0.049 | flat |
| `parse_tree_metadata` | 0.053 / 0.054 / 0.053 | 0.053 / 0.053 / 0.052 | flat |

The two metadata benchmarks that read +11% and +16% in run 1 are flat to the microsecond in run 2, on a run whose control noise floor is a fifth of run 1's. That is the code-layout scatter #204 documented and #240 measured — the gate's 7% threshold sits inside it — and it is why the gate's own output says to re-run before acting on a verdict.

Verdict stands on the agreement of three instruments that do not share the runner: the x86 gate twice, a local interleaved A/B on an M4 twice (`parse_lazy_untouched` 0.041 → 0.030 ms, `parse_tree_lazy_untouched` 0.040 → 0.030 ms, nothing else past the noise floor), and the allocation counts in `bench/tests/allocs.rs`, which are deterministic and need no runner at all.

PR #259, opened 2026-09-17
One MIME-tree traversal instead of two (#237)

### LINKED ISSUE #237: Collapse MimePart::build and build_node into one MIME-tree traversal parameterised on the leaf policy
## Problem

The PyO3-free core holds two MIME-tree walkers that are the same recursive traversal with a different leaf arm: `MimePart::build` (full mode, #99) and `build_node` (lazy/metadata modes, #202). Both check the depth cap, both read `ctype.mimetype` and the disposition, both recurse over `subparts` for `multipart/*`, both contain a verbatim copy of the `message/rfc822` decode -> `repair_missing_separator` -> `parse_mail` -> recurse block, and both end in the same six-field node literal (`content_type`, `headers`, `filename`, `content_id`, `disposition`, `is_message`). The only thing that differs is what a leaf's body becomes: `Option<Vec<u8>>` in one, `NodeBody` in the other. The two node structs differ in exactly that one field as well.

This matters for a library whose reason to exist is a parser that is fast *and* agrees with itself: every per-node rule the tree enforces -- the depth cap, the embedded-message repair, how `Content-ID` is normalised, how the disposition token is derived -- has to be edited in two places, and the project already pays for a fuzz invariant (`parse_agreement` #8) whose only job is to notice when the two copies drift. Making the traversal exist once turns that invariant from the last line of defence into a regression check, and removes the second copy of a security-relevant block (the depth cap and the re-parse of attacker-supplied embedded messages). The existing code also carries a false comment: `build_node` says it is `#[inline(never)]` "like `MimePart::build`", but `MimePart::build` has no such attribute.

Performance is **expected flat**: two monomorphised `#[inline(never)]` recursions replace two hand-written ones, so the instantiation count seen by the linker is unchanged. This is a maintainability change and must be proven not to move the hot path (see CONTRIBUTING "Cold code can slow the hot path").

## Evidence

- `src/mail_parser.rs:908-912` -- `impl MimePart {` / `fn build(part: &ParsedMail<'_>, depth: usize) -> Result<Self, MailParseError> {` / `if depth >= MAX_MIME_DEPTH { return Err(MailParseError::Generic(ERR_MIME_DEPTH)); }`. No `#[inline(never)]` on lines 908-909.
- `src/mail_parser.rs:1070-1080` -- `/// `#[inline(never)]`, like `MimePart::build`: it recurses, it returns a large` / `#[inline(never)]` / `fn build_node(part, depth, defer) -> Result<TreeNode, MailParseError> {` / the same `MAX_MIME_DEPTH` check. The comment's claim about `MimePart::build` is false (grep `inline(never)` in the file: hits at 600, 768, 1072, none near 908).
- `src/mail_parser.rs:914-915` vs `1082-1083` -- identical: `let mime = part.ctype.mimetype.as_str(); let disposition = part.get_content_disposition();`.
- `src/mail_parser.rs:917-926` vs `1085-1091` -- identical `multipart/` arm: `part.subparts.iter().map(|child| <recurse>(child, depth + 1 ...)).collect::<Result<Vec<_>, _>>()?`.
- `src/mail_parser.rs:927-942` vs `1092-1105` -- verbatim copy of the embedded-message arm: `let raw = part.get_body_raw()?; let inner = { let repaired = repair_missing_separator(&raw); let parsed = parse_mail(repaired.as_deref().unwrap_or(raw.as_slice()))?; <recurse>(&parsed, depth + 1 ...)? };`.
- `src/mail_parser.rs:943-945` vs `1106-1116` -- the only real difference: `(Some(part.get_body_raw()?), Vec::new())` vs `NodeBody::Undecoded { encoded_size: encoded_size(part), raw: defer.then(|| part.raw_bytes.to_vec()) }`.
- `src/mail_parser.rs:946-957` vs `1119-1130` -- identical node literal: `content_type: mime.to_string(), headers: collect_headers(part), filename: part_filename(&disposition, &part.ctype), content_id: part.get_headers().get_first_value("Content-ID").map(|raw| normalize_content_id(&raw)), disposition: disposition_token(part, &disposition.disposition), is_message: mime == "message/rfc822"`.
- `src/mail_parser.rs:895-906` (`struct MimePart`) vs `1032-1041` (`struct TreeNode`) -- same six fields plus `children`; differ only in `content: Option<Vec<u8>>` vs `body: NodeBody`.
- `src/mail_parser.rs:1024-1030` -- why the second struct exists: "Deliberately not `MimePart` with a wider `content`. `MimePart.content` is `Option<Vec<u8>>` and its `None` *means* \"this is a container\"; a mode where `None` could also mean \"not decoded yet\" would make the two indistinguishable". A generic body type satisfies this without a second traversal.
- `src/mail_parser.rs:963-971` and `1055-1068` -- the two entry points `parse_email_tree(payload)` and `parse_tree_deferred(payload, defer: bool)`, each doing the same `MAX_INPUT_BYTES` check and `repair_missing_separator` before calling its walker.
- `src/fast_mail_parser.rs:398, 412` -- `fn from_part(py, part: mail_parser::MimePart)` reads `content: part.content`; `606, 620` and `626, 636` -- `metadata_node`/`lazy_node` take `mail_parser::TreeNode` and read `node.body`; `1737` -- `mail_parser::parse_tree_deferred(message.as_ref(), defer)`.
- `fuzz/fuzz_targets/parse_agreement.rs:38-44` -- invariant 8: "A tree is the same tree in all three modes (#202) ... Every per-node field except the body must be identical to full mode's"; `344-364` compare `canonical_full(full)` against `canonical_node(&described)` and `canonical_node(&deferred)`. The fuzz crate includes the core by path (`fuzz/fuzz_targets/parse_agreement.rs:48-50`) and names `mail_parser::MimePart` (54, 60, 116, 122, 136, 168), `.content` (68, 137, 169), `mail_parser::TreeNode` (96, 102, 150, 420), `.body` (153, 421) and `parse_tree_deferred(data, false|true)` (345-346).
- `src/fast_mail_parser.rs:599-604` -- the maintainers' position on the *binding* pair: "Two `#[inline(never)]` recursions rather than one generic ... a generic over that would be two instantiations of the same code with an extra layer to read." That reasoning is about the Py* projections and is respected here: the binding is out of scope.
- `CONTRIBUTING.md:305-318` -- "This crate is unusually sensitive to codegen" and "Cold code can slow the hot path ... adding a third instantiation stopped the one wrapping `parse_email` from being inlined -- costing 24% on large messages while the new code was never executed (#99)". Known profile: full parse of `tests/data/large_message.eml` 0.225 ms (M4) / 0.604 ms (EPYC); the tree walkers are not in the top loops (`decode_base_mut` 53%, whitespace strip 27%, `memmem` 7.5%).

Relates to round-1 issue "Share one envelope/part-classification helper pair across the three flat parsers (metadata, lazy, full)" (the flat-side twin of this; `extract_mail_parts` at `src/mail_parser.rs:1444-1450` carries the third `MAX_MIME_DEPTH` check and is that issue's business) and "Split src/fast_mail_parser.rs into modules and dedupe the copied getters and parse_many result loops" (binding side).

## Proposal

Make the tree node generic over its body and drive one traversal with a small leaf policy. All types stay `pub(crate)`; nothing in the Python surface changes.

```rust
/// One node of the MIME tree; `B` is what a body is in the mode that built it.
#[derive(Debug)]
pub(crate) struct Node<B> {
    pub(crate) content_type: String,
    pub(crate) headers: Vec<(String, Vec<String>)>,
    pub(crate) filename: String,
    pub(crate) content_id: Option<String>,
    pub(crate) disposition: Option<String>,
    pub(crate) is_message: bool,
    pub(crate) body: B,
    pub(crate) children: Vec<Node<B>>,
}

/// Full mode: `None` means container, `Some` means decoded -- unchanged meaning.
pub(crate) type MimePart = Node<Option<Vec<u8>>>;
/// Deferred modes: the three-way `NodeBody` enum, unchanged.
pub(crate) type TreeNode = Node<NodeBody>;

/// What a mode does with a body. Everything else about a node is mode-independent.
trait LeafPolicy {
    type Body;
    fn container(&self) -> Self::Body;
    /// `raw` is the already-decoded `message/rfc822` body the walk had to decode to find the children.
    fn message(&self, part: &ParsedMail<'_>, raw: Vec<u8>) -> Self::Body;
    fn leaf(&self, part: &ParsedMail<'_>) -> Result<Self::Body, MailParseError>;
}

struct Full;
impl LeafPolicy for Full {
    type Body = Option<Vec<u8>>;
    fn container(&self) -> Self::Body { None }
    fn message(&self, _: &ParsedMail<'_>, raw: Vec<u8>) -> Self::Body { Some(raw) }
    fn leaf(&self, part: &ParsedMail<'_>) -> Result<Self::Body, MailParseError> { Ok(Some(part.get_body_raw()?)) }
}

struct Deferred { defer: bool }
impl LeafPolicy for Deferred {
    type Body = NodeBody;
    fn container(&self) -> NodeBody { NodeBody::Container }
    fn message(&self, part: &ParsedMail<'_>, raw: Vec<u8>) -> NodeBody {
        NodeBody::Decoded { encoded_size: encoded_size(part), content: raw }
    }
    fn leaf(&self, part: &ParsedMail<'_>) -> Result<NodeBody, MailParseError> {
        Ok(NodeBody::Undecoded { encoded_size: encoded_size(part), raw: self.defer.then(|| part.raw_bytes.to_vec()) })
    }
}

/// The one traversal. `#[inline(never)]`: it recurses, it returns a large struct, no hot path reaches it.
#[inline(never)]
fn build_node<P: LeafPolicy>(part: &ParsedMail<'_>, depth: usize, policy: &P) -> Result<Node<P::Body>, MailParseError> {
    // depth cap, mime/disposition, multipart arm, message/rfc822 arm, leaf arm, node literal -- once.
}
```

`parse_email_tree` becomes `build_node(&parsed, 0, &Full)`; `parse_tree_deferred(payload, defer)` keeps its signature and calls `build_node(&parsed, 0, &Deferred { defer })`. The two entry points' shared prologue (`MAX_INPUT_BYTES` check + `repair_missing_separator`) may be folded into one private `parse_tree_with(payload, &policy)` that both call.

Behaviour that must stay identical: error text and ordering (`ERR_MIME_DEPTH` at the same depth, `ERR_INPUT_TOO_LARGE`), the `message/rfc822` repair-then-reparse with the nesting counted against the same cap, `Decoded` for embedded messages in both deferred modes, `Undecoded { raw: None }` in metadata mode and `Some(raw_bytes)` in lazy mode, `None`-means-container in full mode, header order (#157), and the exact per-node field derivations. Two monomorphisations (`Full`, `Deferred`) -- the same number of compiled recursions as today.

New dependency: none. Vendored mailparse: untouched.

1. `src/mail_parser.rs`: replace `struct MimePart` (895-906), `impl MimePart { fn build }` (908-959), `struct TreeNode` (1032-1041) and `fn build_node` (1070-1132) with `Node<B>`, the two type aliases, `LeafPolicy`, `Full`, `Deferred` and the single generic `build_node`. Keep `NodeBody` (983-1022) and `NodeBody::encoded_size` as they are. Keep the doc comments' substance -- move the "container's body is its children" note and the `message/rfc822` rationale (929-938, 1093-1094) onto the single traversal, and rewrite the 1024-1030 paragraph to say the type parameter is what keeps "container" and "not decoded yet" distinguishable. Fix the stale `#[inline(never)]` comment by making it true (the one traversal carries the attribute).
2. Field rename: `MimePart.content` becomes `.body` (one name for the generic field). Update `src/fast_mail_parser.rs:412` (`content: part.content` -> `content: part.body`) and `fuzz/fuzz_targets/parse_agreement.rs:68, 137, 169` (`part.content` -> `part.body`). Nothing else in the binding or fuzz target should need to change if the aliases keep the names `MimePart`, `TreeNode` and the `parse_tree_deferred(payload, bool)` signature.
3. Do not touch `src/fast_mail_parser.rs` beyond step 2. The three Py* projections (`from_part` 398-415, `metadata_node` 606-623, `lazy_node` 626-664) are deliberately left alone: the maintainers declined a generic there (599-604) and they are public `#[pyclass]` types.
4. The fuzz crate compiles the core by path (`fuzz/Cargo.toml`, `#[path = "../../src/mail_parser.rs"]`). It must still build: `cd fuzz && RUSTUP_TOOLCHAIN=nightly cargo fuzz build` (needs `cargo install cargo-fuzz --locked`). CI's 60-second short pass (`.github/workflows/test.yml`, job `fuzz`) will fail the PR otherwise.
5. Oracles: `tests/test_mime_tree.py` (all; especially `test__an_embedded_message_is_parsed_not_opaque`, `test__a_doubly_nested_message_keeps_both_levels`, `test__embedded_message_nesting_counts_against_the_recursion_cap`, `test__leaf_content_agrees_with_the_flat_api`) and `tests/test_tree_modes.py` (all; especially `test__the_shape_is_identical_to_full_mode`, `test__a_leafs_bytes_are_identical_to_full_mode`, `test__encoded_size_is_present_exactly_where_content_is`, `test__a_single_part_message_defers_its_root`, `test__an_embedded_message_arrives_already_decoded`, `test__metadata_mode_can_raise_on_a_broken_embedded_message`, `test__the_mime_depth_cap_holds`, `test__the_input_size_cap_holds`). `tests/test_stdlib_parity.py` and `tests/test_contract.py` must pass unchanged -- no `DIVERGENCES` edit and no frozen-set edit is expected, because nothing public moves.
6. Traps: never run cargo inside `vendor/mailparse` without `--target-dir` elsewhere; `vendor/` is not touched so `PATCH.md` needs no update; abi3 (`abi3-py311`) is unaffected; if you split the tree code into a new file, put it directly under `src/` -- the sdist include glob is `src/*` (`pyproject.toml:46-49`), not recursive; `__init__.pyi` and `tests/test_contract.py` must NOT change (if they do, the change has leaked into the public API and is wrong); clippy runs with `-D warnings -W clippy::cast_possible_truncation`, so a `&Full` ZST argument must not trigger `needless_pass_by_ref`-style lints -- pass policies by reference consistently or take them by value, but keep two instantiations.
7. Add a `### Changed` entry under `## [Unreleased]` in `CHANGELOG.md` (Keep-a-Changelog format, one bullet naming #99/#202 and stating one traversal, no behaviour change).

## Harness -- how to prove it

Build both sides in two venvs from two worktrees (`git worktree add ../fmp-base master`), each: `uv venv && uv pip install maturin pytest pytest-benchmark && maturin develop --release`.

Correctness (branch venv):

```
cargo fmt --all -- --check
ruff check .
cd fuzz && mkdir -p corpus/parse_agreement && cp ../tests/data/*.eml ../tests/data/rfc/*.eml corpus/parse_agreement/ \
  && RUSTUP_TOOLCHAIN=nightly cargo fuzz run parse_agreement -- -max_total_time=60 -max_len=65536
```

The fuzz run is the invariant-8 oracle: it must finish with zero crashes (a shape disagreement panics with "the metadata tree's shape disagrees with full mode" or "the lazy tree's shape disagrees with full mode").

Performance (interleaved, 3 rounds A/B/A/B/A/B, alternating venvs):

```
for r in 1 2 3; do
  (source ../fmp-base/.venv/bin/activate && pytest tests/benchmark/test_read_message.py --benchmark-only --benchmark-json=/tmp/a$r.json)
  (source .venv/bin/activate          && pytest tests/benchmark/test_read_message.py --benchmark-only --benchmark-json=/tmp/b$r.json)
done
AB_THRESHOLD_PCT=7 python .github/scripts/ab_median.py --a-label master --b-label branch \
  --a /tmp/a1.json /tmp/a2.json /tmp/a3.json --b /tmp/b1.json /tmp/b2.json /tmp/b3.json
```

Decision rule: no benchmark may regress more than 7% (the CI gate's `AB_THRESHOLD_PCT: "7"`, `.github/workflows/test.yml:460`) once the pure-Python controls' spread is subtracted. Expected to be flat (within the control floor): `test__fast_mail_parser___parse_tree`, `test__fast_mail_parser___parse_tree_metadata`, `test__fast_mail_parser___parse_tree_lazy_untouched`. Must also be flat because the change is cold code that can still move the hot path via layout (#99, CONTRIBUTING 311-318): `test__fast_mail_parser___parse_message`, `test__fast_mail_parser___parse_metadata`, `test__fast_mail_parser___parse_lazy_untouched`, `test__fast_mail_parser___parse_many`. No benchmark is expected to improve; a claimed improvement should be treated as layout noise, not a result.

No new benchmark is needed. Add one new Rust unit test in `src/mail_parser.rs` (`#[cfg(test)]`) that builds the same fixture (`tests/data/large_message.eml` read at compile time or a small inline multipart with a nested `message/rfc822`) with `Full` and with `Deferred { defer: false }` and `Deferred { defer: true }`, and asserts the three trees render identically on every non-body field -- the in-tree, always-run twin of fuzz invariant 8.

Put the local before/after `ab_median.py` table (with the CPU line it prints) in the PR body. The M4 is not a proxy for x86: the codegen-layout swing (#204, up to 96% on Zen) does not reproduce on the M4, so the CI gate on the x86 runner is the verdict; if it reports a regression the change "cannot possibly" cause, re-run it before bisecting (CONTRIBUTING 355-363) and, if it reproduces, compare `toolchain-ab.yml`-style on the same runner class.

## Acceptance criteria

- [ ] `src/mail_parser.rs` contains exactly one recursive MIME-tree builder; `grep -c 'if depth >= MAX_MIME_DEPTH' src/mail_parser.rs` returns 2 (the tree builder and the flat `extract_mail_parts`), down from 3.
- [ ] The `message/rfc822` decode -> `repair_missing_separator` -> `parse_mail` -> recurse block appears once in the tree code.
- [ ] `MimePart` and `TreeNode` remain the names the binding and fuzz target use; `parse_email_tree(payload)` and `parse_tree_deferred(payload, bool)` keep their signatures.
- [ ] The `#[inline(never)]` comment on the traversal is true of the function it describes.
- [ ] `src/fast_mail_parser.rs` changes are limited to the `part.content` -> `part.body` rename at the `from_part` site; `fast_mail_parser/__init__.pyi`, `tests/test_contract.py`, `docs/compatibility.md` and `vendor/` are unchanged.
- [ ] `pytest tests --ignore=tests/benchmark` passes, including all of `tests/test_mime_tree.py` and `tests/test_tree_modes.py`, with no `DIVERGENCES` edits in `tests/test_stdlib_parity.py`.
- [ ] `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings -W clippy::cast_possible_truncation`, `mypy --strict fast_mail_parser/`, `ruff check .` and the vendored mailparse tests (with an external `--target-dir`) pass.
- [ ] `cd fuzz && RUSTUP_TOOLCHAIN=nightly cargo fuzz build` succeeds and a 60 s `cargo fuzz run parse_agreement` reports zero crashes; CI's `fuzz (short pass)` job is green.
- [ ] New `#[cfg(test)]` unit test asserts the three policies produce identical non-body node fields on a nested `message/rfc822` fixture.
- [ ] Interleaved local A/B (3 rounds, `ab_median.py`, threshold 7%) shows no treatment benchmark regressed beyond the control floor; table and CPU line are in the PR body; the CI benchmark gate on x86 is green.
- [ ] `CHANGELOG.md` has an `[Unreleased]` `### Changed` entry.

## Out of scope

- Deduplicating the three binding projections `PyMimePart::from_part`, `metadata_node`, `lazy_node` -- the maintainers' recorded position at `src/fast_mail_parser.rs:599-604`; revisit only if the round-1 "Split src/fast_mail_parser.rs into modules" issue changes that.
- The flat parsers' shared classification helpers (`extract_mail_parts` and the three flat entry points) -- round-1 issue "Share one envelope/part-classification helper pair across the three flat parsers".
- Any change to `NodeBody`'s variants, to the lazy-mode `raw_bytes` retention, or to what any mode decodes.
- Any public API, `.pyi`, or `mode=` semantics change.
- Performance work in the tree path (it is not in the measured profile).

## Related issues

- #234
- #239 — `build_node` gains a `base` parameter there; do this collapse first or together

kurok:
Closes #237.

> **Stacked on #258** (`feat/239-lazy-borrow`), so the diff here is only this change. #237's own note says *"#239 — `build_node` gains a `base` parameter there; do this collapse first or together"*; #239 went first, and its `Retain` enum turns out to be exactly the leaf policy this issue asks the traversal to be parameterised on. Merge #258 first and this retargets to master on its own.

## The duplication

`MimePart::build` (full mode, #99) and `build_node` (the deferred modes, #202) were the same recursive walk with a different leaf arm — same depth cap, same `multipart/*` recursion, a verbatim copy of the `message/rfc822` decode → repair → re-parse block, and the same six-field node literal. Every per-node rule had to be edited twice, and a fuzz invariant existed to notice when the copies drifted.

One of those copies was of **the depth cap and the re-parse of an attacker-supplied embedded message**. That is the part worth not having two of.

## The shape

```rust
pub struct Node<B> { content_type, headers, filename, content_id, disposition, is_message, body: B, children: Vec<Node<B>> }
pub type MimePart = Node<Option<Vec<u8>>>;
pub type TreeNode = Node<NodeBody>;

trait LeafPolicy: Copy {
    type Body;
    fn inside_embedded(self) -> Self;   // the subtree's bytes are in no caller buffer
    fn container(self) -> Self::Body;
    fn embedded(self, part: &ParsedMail<'_>, raw: Vec<u8>) -> Self::Body;
    fn leaf(self, part: &ParsedMail<'_>) -> Result<Self::Body, MailParseError>;
}
```

Two implementations, so the linker sees the two instantiations it saw before: `Full`, and `Retain` from #239. `inside_embedded` is a hook rather than a constant precisely because of #239 — a mode that retains offsets has to stop retaining them inside an embedded message, whose bytes are a decode of the enclosing body.

`NodeBody` stays, and so does its reason: `MimePart`'s `None` *means* container, and a mode where `None` could also mean "not decoded yet" would make the two indistinguishable. Generic over the body satisfies that without a second walk.

## Acceptance

| | |
| --- | --- |
| `grep -c 'if depth >= MAX_MIME_DEPTH'` in the core | **3 → 2** (traversal + flat `extract_mail_parts`) |
| `message/rfc822` decode→repair→re-parse block | once |
| `MimePart` / `TreeNode` names, `parse_email_tree(payload)`, `parse_tree_deferred(payload, bool)` | unchanged¹ |
| `src/fast_mail_parser.rs` | 1 line (`part.content` → `part.body`) |
| `__init__.pyi`, `tests/test_contract.py`, `docs/`, `vendor/` | untouched |

¹ `parse_tree_deferred` returns `DeferredTree` rather than `TreeNode`, which is #258's change, not this one.

The false comment is fixed too: `build_node` claimed to be `#[inline(never)]` "like `MimePart::build`", which had no such attribute.

## Measurements

Local interleaved A/B on an M4 (`Apple M4, 10 vCPU`), 3 rounds, against this PR's base:

| Benchmark | base (#258) | this PR | |
| --- | --- | --- | --- |
| `parse_tree` | 0.156 ms | 0.159 ms | −2.0% |
| `parse_tree_metadata` | 0.030 ms | 0.030 ms | −0.3% |
| `parse_tree_lazy_untouched` | 0.030 ms | 0.030 ms | −0.3% |
| `parse_message` | 0.156 ms | 0.158 ms | −1.0% |
| `parse_metadata` | 0.029 ms | 0.029 ms | −0.9% |
| `parse_lazy_untouched` | 0.030 ms | 0.031 ms | −1.6% |
| `parse_many` | 1.319 ms | 1.322 ms | −0.2% |

Noise floor from the pure-Python controls: **1.8%**. Worst real movement −2.0% (`parse_tree`), whose per-round values are 0.156 / 0.156 / 0.155 against 0.156 / 0.159 / 0.162 — identical in round 1 and drifting upward after, which is thermal, not codegen. Flat, as a maintenance change should be; the x86 gate is the verdict.

`ab_median.py` also flags `attachment_reread` at "+97.6%" **in this PR's favour**. Ignore it: that benchmark is 42–83 ns and both sides sampled the same two timer ticks (base 83/42/83, this PR 83/42/42). It is quantisation, not a measurement.

**Verification**: 864 Python tests including all of `test_mime_tree.py` and `test_tree_modes.py` with no `DIVERGENCES` edits; 11 core Rust tests; `cargo clippy --all-targets -D warnings -W clippy::cast_possible_truncation` on all three crates; `cargo fmt`; `mypy --strict`; `ruff`; the vendored mailparse suite (82 tests, external `--target-dir`); `check_vendored_mailparse.sh` and `check_bench_lockfile.py`. 338K `cargo fuzz run -a parse_agreement` executions, zero crashes — that target's invariant 8 is the shape oracle this change has to satisfy, and it now also has an always-run twin in `cargo test` (`every_mode_builds_the_same_tree_around_an_embedded_message`).


kurok:
### CI status, and the gate's scatter on the small benchmarks

**11 of 11 registered checks pass.** The four that did not register are CodeQL's (`Analyze (actions|python|rust)` and `CodeQL`): this repo uses GitHub's *default setup* for code scanning (`dynamic/github-code-scanning/codeql`), which only runs on pull requests targeting the default branch. They will run on their own once #258 merges and this PR retargets to master. Nothing to do about it here, but worth saying out loud rather than letting "11 green" read as "fully gated".

**The benchmark gate failed once and passed on re-run**, on the same five benchmarks that did this on #258. There are now five measurements of this code against master, on four different CPU models:

| | runner | control floor | `parse_metadata` / `_str` / `parse_many_metadata` / `parse_tree_metadata` | verdict |
| --- | --- | --- | --- | --- |
| #258 run 1 | Xeon 8370C | 3.9% | +11.2 / +11.2 / +12.7 / +16.3% | fail |
| #258 run 2 | Xeon 8370C | 0.8% | flat | pass |
| #258 run 3 | EPYC 7763 | 0.3% | +0.3 / +0.5 / +0.1 / −0.2% | pass |
| #259 run 1 | EPYC 9V45 | 2.2% | +10.0 / +9.8 / +9.9 / +11.8% | fail |
| #259 run 2 | EPYC 9V74 | 6.5% | +0.8 / +0.8 / +0.0 / +2.2% | pass |

Two things to read off it.

**These four benchmarks carry roughly ±10% run-to-run scatter that the build does not explain.** `parse_metadata` and `parse_metadata_str` go through `parse_email_metadata` → `metadata_from_payload`, which neither this PR nor #258 modifies, and `parse_many_metadata` is `parse_many_as(..., parse_email_metadata)`. They are also the smallest gated benchmarks — 0.028–0.051 ms depending on runner. The verdict flips on the same binaries, and it does not track the control noise floor either (run 5 has the highest floor of the five and is flat). This is the scatter #240 exists to measure, and the gate's 7% threshold sits inside it.

**What is consistent is the thing that is real.** `parse_lazy_untouched` and `parse_tree_lazy_untouched` improve in every one of the five runs, on every CPU model: −19.1%, −23.0%, −29.7%, −32.1%, −32.9% and −20.9%, −18.7%, −31.4%, −32.9%, −31.4%. A code-layout artefact does not reproduce five times across four CPU models; a removed copy does. (That improvement is #258's, not this PR's — the gate resolves a stacked PR's base to `origin/master`, so these runs measure #239 and #237 together.)

**For this PR's own effect**, the comparison that isolates it is the local interleaved A/B against its actual base in the PR body: worst real movement −2.0% (`parse_tree`) against a 1.8% control floor, on an M4. Flat, which is what a collapse of two cold recursions into one generic should be.

PR #260, opened 2026-09-17
One envelope reader and one part classifier across the flat parsers (#234)

### LINKED ISSUE #234: Share one envelope/part-classification helper pair across the three flat parsers (metadata, lazy, full)
## Problem

`src/mail_parser.rs` has three flat entry points -- `metadata_from_payload`, `lazy_from_payload` and `Mail::from_payload` -- and each carries its own copy of two blocks: the 8-field envelope extraction (headers map, Subject, Date, From/To/Cc/Bcc/Reply-To) and the per-part classification (skip `multipart/*`, resolve disposition and filename, apply the RFC 2183 body-vs-attachment rule, normalise Content-ID, derive the disposition token). The two MIME-tree builders carry a fourth and fifth copy of the filename/Content-ID/disposition derivation. The copies are already drifting in small ways (the `header_addresses` doc comment counts ten call sites; there are fifteen), and a fuzz target (`parse_agreement`, invariants 3 and 7) exists specifically to catch them disagreeing about a message.

The duplication is documented as deliberate, but the recorded objection is to *threading a runtime mode flag* through the hot parse (`#100` measured +47% for one bool and one branch). A branch-free, straight-line helper that every mode calls -- the same instructions each call site emits today -- is not that. The maintainer already applies this principle in the tree API (`NodeBody` at 976-981: "Splitting the struct would have split that derivation too, which is the thing the modes must not disagree about"); the flat modes should get the same treatment.

This is maintainability work. It moves no hot loop: the known full-parse profile is 53% `data_encoding::decode_base_mut`, 27% the base64 whitespace strip, 7.5% `memmem`, all inside `vendor/mailparse` and its dependencies, which this change does not touch. Expected performance delta is 0%; what it buys is one place to add an envelope header (e.g. a future `Message-ID`) or change a classification rule instead of three to five.

## Evidence

All citations are `src/mail_parser.rs` at master `73f56da` unless stated.

- **Envelope copy 1** -- 603-628, `metadata_from_payload`:
  ```rust
  let headers = collect_headers(&mail);
  let subject = mail.get_headers().get_first_value("Subject").unwrap_or_default();
  let date = mail.get_headers().get_first_value("Date").unwrap_or_default();
  ...
  let from_list = header_addresses(&mail, "From", &mut discarded);
  let from_ = from_list.into_iter().next();
  let to = header_addresses(&mail, "To", &mut discarded);
  let cc = header_addresses(&mail, "Cc", &mut discarded);
  let bcc = header_addresses(&mail, "Bcc", &mut discarded);
  let reply_to = header_addresses(&mail, "Reply-To", &mut discarded);
  ```
- **Envelope copy 2** -- 773-793, `lazy_from_payload`: the same eleven statements with `&mut warnings`, followed by `if !date.is_empty() && parse_date_epoch(&date).is_none() { warn_date(&mut warnings, &date); }` (791-793).
- **Envelope copy 3** -- 1306-1338, `Mail::from_payload`: identical to copy 2 including the `warn_date` check (1336-1338).
- **Classification copy 1** -- 633-660 (metadata): `if mime.starts_with("multipart/") { continue; }`, `let disposition = part.get_content_disposition();`, `let is_body = disposition.disposition != DispositionType::Attachment && matches!(mime, "text/plain" | "text/html");`, then `filename: part_filename(&disposition, &part.ctype)`, `content_id: part.get_headers().get_first_value("Content-ID").map(|raw| normalize_content_id(&raw))`, `disposition: disposition_token(&part, &disposition.disposition)`.
- **Classification copy 2** -- 800-841 (lazy) and **copy 3** -- 1346-1397 (full): the same block. Only the attachment payload differs: `encoded_size` alone (659), `encoded_size` + `raw: part.raw_bytes.to_vec()` (835-839), `content` (1391-1397).
- **Copies 4 and 5 of the part-identity derivation** -- 948-954 (`MimePart::build`) and 1121-1127 (`build_node`): `filename: part_filename(&disposition, &part.ctype)`, `content_id: part.get_headers().get_first_value("Content-ID").map(|raw| normalize_content_id(&raw))`, `disposition: disposition_token(part, &disposition.disposition)`.
- **Drift already present** -- 1225: `/// A thin wrapper so its ten call sites stay one short line each`. Actual call sites: 623-628, 784-789, 1324-1329 = fifteen. The comment predates lazy mode (#198) and was not updated when the third copy landed.
- **Near-duplicate structs** -- `LazyMail` 716-729 vs `Mail` 1238-1254: twelve identical fields except `attachments: Vec<LazyAttachment>` vs `Vec<Attachment>`. `LazyAttachment` 698-707 is `Attachment` 1257-1263 with `content` replaced by `encoded_size` + `raw`.
- **Rationale for the duplication** -- 572-576: "The envelope extraction below repeats `Mail::new`'s, deliberately: both are only calls into the shared helpers ... Threading a mode through `Mail::new` instead would have put a branch in the hot path for the benefit of the cold one." And 747-751: "threading a mode through `Mail::from_payload` instead is what cost the hot path 47% in #100." Both reject a runtime branch; neither covers a shared straight-line helper.
- **The one-call-site constraint a refactor must keep** -- 1284-1287: "`from_payload` carries no `inline(never)`, so a second call site is a second chance to inline the whole parse body, and duplicating it there cost the flat path 28%."
- **Drift detector that exists because of the copies** -- `fuzz/fuzz_targets/parse_agreement.rs:17-21`: "**The envelope agrees.** Subject, date and the whole header map must be identical between the two modes." and 31-37: "The envelope, the bodies and the whole warning list must agree too".
- **Codegen caution that still applies** -- `git show fd12220` (#188): the #100 restructure left metadata mode at +96% "because it was still routed through a new shared slow-path helper"; `CHANGELOG.md` 0.9.0: "The version bump itself moved the decoding paths +3-7% against the last master commit on the M4". A shared helper here is *expected* to be free, but the gate decides, not the expectation.

## Proposal

Two steps. Step 1 is the issue; step 2 is optional and may be split into a follow-up PR if step 1's A/B is clean.

**Step 1 -- shared helpers, no struct-shape change, no generics.** New dependency: none.

```rust
/// The eight envelope fields every flat mode reads, in one place (#<this issue>).
struct Envelope {
    subject: String,
    date: String,
    from_: Option<Address>,
    to: Vec<Address>,
    cc: Vec<Address>,
    bcc: Vec<Address>,
    reply_to: Vec<Address>,
    headers: Vec<(String, Vec<String>)>,
}

/// Straight-line, no mode parameter: the branch #100 measured is not here.
#[inline]
fn envelope(mail: &ParsedMail<'_>, warnings: &mut Vec<Warning>) -> Envelope { /* the eleven statements, once */ }

/// Everything the three flat loops decide about a part before touching its body.
struct PartInfo {
    filename: String,
    content_id: Option<String>,
    disposition: Option<String>,
    is_body: bool,
}

/// `None` for a `multipart/*` container (#22). `is_body` is the RFC 2183 rule (#25).
#[inline]
fn classify_part(part: &ParsedMail<'_>) -> Option<PartInfo> { ... }
```

Each of the three flat parsers becomes `let Envelope { subject, date, from_, to, cc, bcc, reply_to, headers } = envelope(&mail, &mut warnings);` (metadata passes its existing `discarded`), keeps its own `warn_date` check (so metadata mode does not gain a `dateparse` call -- see the comment at 614-622 for why metadata drops warnings), and in the loop does `let Some(info) = classify_part(&part) else { continue };` then `let mime = part.ctype.mimetype.as_str();`. `MimePart::build` and `build_node` call a `part_identity(part, &disposition) -> (String, Option<String>, Option<String>)` (or `classify_part` and ignore `is_body`) for their filename/content_id/disposition triple. Destructuring keeps `Mail`, `LazyMail`, `MailMetadata` field shapes unchanged, so `src/fast_mail_parser.rs` (`from_mail` 882-910, `from_lazy` 1137-1164, `from_metadata` 305-322) and the fuzz `canonical()` renderers (`fuzz/fuzz_targets/parse_email.rs:35-62`) need no edits.

**Step 2 (optional) -- one loop for full and lazy.** Replace `Mail` + `LazyMail` by a single `FlatMail<A>` with `type Mail = FlatMail<Attachment>; type LazyMail = FlatMail<LazyAttachment>;` (aliases keep the binding layer and fuzz targets unchanged), and

```rust
trait PartSink: Sized {
    fn from_part(part: &ParsedMail<'_>, mime: &str, info: PartInfo) -> Result<Self, MailParseError>;
}
impl PartSink for Attachment     { /* content: part.get_body_raw()? */ }
impl PartSink for LazyAttachment { /* encoded_size: encoded_size(part), raw: part.raw_bytes.to_vec() */ }

fn flat_from_payload<A: PartSink>(payload: &[u8]) -> Result<FlatMail<A>, MailParseError> { /* today's loop, once */ }
```

with `Mail::new` calling `flat_from_payload::<Attachment>` and `parse_email_lazy` calling `flat_from_payload::<LazyAttachment>` -- exactly one call site per instantiation, preserving the property recorded at 1284-1287. Monomorphisation yields one function body per mode, as today.

**Behaviour that must stay identical (all modes):**
- Full mode decodes every non-multipart part (`get_body_raw()?` at 1375 runs for attachments too), so an attachment with a broken transfer encoding still raises `ParseError` in full mode and does not in lazy/metadata (documented difference, `parse_agreement` invariant 2). With `PartSink`, `Attachment::from_part` must call `get_body_raw()?`.
- Metadata mode collects warnings into a discarded list and never calls `parse_date_epoch` (614-628).
- Warning order and `part_path` indices (`attachments.len()`, `text_html.len()`, `text_plain.len()` at the moment of push) are unchanged; `strict=True` reports the first warning.
- `header_addresses` reads the *first* occurrence of each header; `from_` takes the first mailbox.
- Header map key order (#157) is `collect_headers`'s and is untouched.

1. Read `CONTRIBUTING.md` lines 295-365 (Performance) first. Note "do not assume a change is free because it looks free" and the re-run rule for a large gate failure.
2. In `src/mail_parser.rs`, add `Envelope`, `envelope()`, `PartInfo`, `classify_part()` next to the existing helpers (`part_filename` 447, `normalize_content_id` 460, `disposition_token` 473, `collect_headers` 501, `encoded_size` 528, `header_addresses` 1229). Do **not** give either helper a mode/bool parameter.
3. Replace the three envelope blocks (603-628, 773-793, 1306-1338) with a destructuring call. Keep the two `warn_date` checks (791-793, 1336-1338) at their callers.
4. Replace the three classification blocks (633-660, 800-841, 1346-1397) with `classify_part`; keep each mode's attachment payload and body handling in place. Reuse the identity triple in `MimePart::build` (948-954) and `build_node` (1121-1127).
5. Rewrite the rationale comments at 572-576 and 747-751 so they describe what is now true (shared straight-line helpers; the branch-free property is what keeps them off the #100 path), and fix the "ten call sites" comment at 1225.
6. Optional step 2 as sketched. Keep `#[inline(never)]` on the thin wrappers that hold the repair (`Mail::new`, `parse_email_lazy`, `parse_email_metadata`) exactly as today; do not add a second call site to any instantiation.
7. Oracles: `tests/test_metadata_mode.py`, `tests/test_lazy_mode.py`, `tests/test_parse_warnings.py`, `tests/test_attachments.py`, `tests/test_mime_tree.py`, `tests/test_tree_modes.py`, `tests/test_stdlib_parity.py` (its `DIVERGENCES` dict at line 45 must not need a new entry), `tests/test_contract.py` frozen attribute sets (must not change). The fuzz targets compile the core by `#[path]` (`fuzz/Cargo.toml`), so `RUSTUP_TOOLCHAIN=nightly cargo fuzz build` and a 30 s `cargo fuzz run parse_agreement` are part of the check -- invariants 3, 4 and 7 are exactly the drift this removes.
8. Traps: `vendor/` is not touched (so `PATCH.md` needs no update and the vendored tests are optional); never run cargo inside `vendor/mailparse` without `--target-dir` elsewhere (the sdist include glob sweeps `target/` in); the Python API does not move, so `fast_mail_parser/__init__.pyi`, `docs/compatibility.md` and the `test_contract.py` frozen sets stay as they are; add a `### Changed` entry under `## [Unreleased]` in `CHANGELOG.md` stating the refactor and the measured A/B result.

## Harness -- how to prove it

Build both sides once, then measure interleaved (the pattern `test.yml` 387-475 uses):

```bash
# two venvs, two wheels
git worktree add /tmp/fmp-base master
python -m venv /tmp/venv-base && /tmp/venv-base/bin/pip install -q '/tmp/fmp-base[test]'
python -m venv /tmp/venv-head && /tmp/venv-head/bin/pip install -q '.[test]'

# interleaved rounds: A B A B A B
for r in 1 2 3; do
  for side in head base; do
    /tmp/venv-$side/bin/pytest -q tests/benchmark --benchmark-min-rounds=25 --benchmark-json=$side-$r.json
  done
done
python .github/scripts/check_benchmark.py head-1.json
```

Benchmarks to read in `tests/benchmark/test_read_message.py`: `test__fast_mail_parser___parse_message` (gate), `test__fast_mail_parser___parse_metadata`, `test__fast_mail_parser___parse_lazy_untouched`, `test__fast_mail_parser___parse_lazy_all_attachments`, `test__fast_mail_parser___parse_many`, `test__fast_mail_parser___parse_many_metadata`, `test__fast_mail_parser___parse_tree`, `test__fast_mail_parser___parse_tree_metadata`, `test__fast_mail_parser___parse_tree_lazy_untouched`. **Expectation: every one within the pure-Python noise floor (about +/-1%); nothing is expected to improve.** No benchmark may regress more than 7%. No new benchmark is needed; add a Rust unit test in `src/mail_parser.rs` `mod tests` asserting `classify_part` returns `None` for a `multipart/mixed` part and `is_body == false` for a `text/plain; Content-Disposition: attachment` part, so the rule has a direct test rather than only the three modes' end-to-end ones.

Full local surface:

```bash
cargo fmt --all -- --check
ruff check .
(cd fuzz && RUSTUP_TOOLCHAIN=nightly cargo fuzz build && \
  mkdir -p corpus/parse_agreement && cp ../tests/data/*.eml ../tests/data/rfc/*.eml corpus/parse_agreement/ && \
  RUSTUP_TOOLCHAIN=nightly cargo fuzz run parse_agreement -- -max_total_time=30 -seed=1)
```

Put the local before/after `ab_median.py` table in the PR body. The Apple M4 is **not** a proxy for the CI runner's x86 Zen: layout effects the M4 cannot see have measured up to +96% there (CONTRIBUTING 324-345), and a version bump alone moved the M4 +3-7% (CHANGELOG 0.9.0). If the PR gate reports a regression above 7% on a change whose local A/B was flat, re-run the gate once before treating it as real; a genuine regression reproduces, a layout artifact does not. `.github/workflows/toolchain-ab.yml` (`workflow_dispatch`) is the interleaved x86 harness to reach for if the verdict is still unclear.

## Acceptance criteria

- [ ] `src/mail_parser.rs` has exactly one place that lists the envelope headers (Subject, Date, From, To, Cc, Bcc, Reply-To, header map) and exactly one that applies the `multipart/*` skip and the RFC 2183 `is_body` rule; `grep -c 'header_addresses(&mail' src/mail_parser.rs` returns 5 (was 15) and `grep -c 'normalize_content_id(&raw)' src/mail_parser.rs` returns 1 (was 5).
- [ ] Neither new helper takes a mode flag, `bool`, or enum selecting behaviour; `Mail::new`, `parse_email_lazy`, `parse_email_metadata` still each contain exactly one call to their parse body.
- [ ] `pytest tests --ignore=tests/benchmark` passes with no change to `tests/test_stdlib_parity.py` `DIVERGENCES` and no change to `tests/test_contract.py`.
- [ ] `cargo fuzz build` succeeds and a 30 s `cargo fuzz run parse_agreement` finds nothing.
- [ ] Interleaved local `ab_median.py` (3 rounds) shows every `test__fast_mail_parser___*` benchmark within the noise floor; the PR gate on x86 passes its 7% threshold (re-run once if it fails on a locally flat change, per CONTRIBUTING).
- [ ] `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings -W clippy::cast_possible_truncation`, `mypy --strict fast_mail_parser/`, `ruff check .` all clean.
- [ ] The rationale comments at 572-576, 747-751 and 1225 are rewritten to match the code; `CHANGELOG.md` `[Unreleased]` has a `### Changed` entry quoting the measured A/B.
- [ ] No file under `vendor/`, `fast_mail_parser/__init__.pyi`, or `docs/` changes.

## Out of scope

- Consolidating the three size-check + `repair_missing_separator` entry-point wrappers (585-600, 752-767, 1272-1291): that would need a generic/closure in the hot path, which is the shape #100 and #187 measured as costly. The #150 metadata miss was in that wrapper, not in the envelope block, and this issue does not claim to prevent a repeat of it.
- Any change to `vendor/mailparse` or to the base64/whitespace loops (the profile's 53% + 27%).
- Public API changes, new modes, new warning kinds, or a metadata-mode warnings channel (614-622 explains why that is a separate decision).
- Merging `MimePart`/`TreeNode` or their builders beyond reusing the part-identity helper.

## Related issues

- #233
- #237 — the tree-side dedupe
- #230

kurok:
Closes #234 (step 1; step 2 deliberately deferred — see below).

> **Stacked on #259**, which is stacked on #258. Each retargets itself as the one below merges.

## The duplication

`Mail::from_payload`, `lazy_from_payload` and `metadata_from_payload` each carried their own copy of the eleven-statement envelope extraction and of the per-part classification. The MIME-tree traversal carried a fourth copy of the identity derivation.

The copies had drifted in the way duplication drifts — `header_addresses` documented *ten* call sites when there were fifteen — and the drift that matters here is not cosmetic. The envelope is what `strict=True` and the whole warning channel are about; the classification rule decides what an attachment **is**. Three derivations of those are three chances for two views of one message to disagree, which is what `parse_agreement`'s invariants 3 and 7 exist to catch.

## The shape

```rust
struct Envelope { headers, subject, date, from_, to, cc, bcc, reply_to }
#[inline] fn envelope(mail: &ParsedMail<'_>, warnings: &mut Vec<Warning>) -> Envelope;

struct PartInfo<'p> { mime: &'p str, disposition: ParsedContentDisposition, is_body: bool }
#[inline] fn classify_part<'p>(part: &'p ParsedMail<'_>) -> Option<PartInfo<'p>>;   // None = multipart/*

struct PartIdentity { filename, content_id, disposition }
#[inline] fn part_identity(part: &ParsedMail<'_>, disposition: &ParsedContentDisposition) -> PartIdentity;
```

Straight-line and `#[inline]`, so each caller emits the instructions it emitted when it owned a copy. What #100 measured at +47% was threading a runtime **mode** through the parse — a branch in the hot path taken for the benefit of the cold one. There is no mode and no branch here, which is the property that makes sharing free.

Counted in the core:

| | before | after |
| --- | --- | --- |
| `DispositionType::Attachment` (the RFC 2183 rule) | 3 | **1**¹ |
| `get_first_value("Content-ID")` derivations | 4 | **1** |
| `starts_with("multipart/")` skips | 3 | **1**² |

¹ the remaining one is `disposition_token`'s own match. ² the other is `extract_mail_parts`'s, which is structural recursion, not classification.

The `warn_date` check stays at the two callers that want it: metadata mode deliberately does not make it, and a parameter to say so would be exactly the branch this avoids.

## One deviation from the issue's sketch

It proposes a single `PartInfo { filename, content_id, disposition, is_body }`. Bundling the rule with the identity would derive a `Content-ID` and a disposition token for **every `text/plain` body in every message** — work no mode does today. So the rule is separate from the identity, and the identity is read only where it is used. A side effect: body parts stop deriving a filename they never used, which full and lazy mode both did.

## Step 2 is not here

The issue's step 2 — `FlatMail<A>` plus a `PartSink` trait, to share the loop body itself — is marked optional and splittable "if step 1's A/B is clean". It is clean, and step 2 rewrites `Mail::from_payload`, the hottest function in the crate. It deserves its own measurement rather than riding along with this one. Happy to do it as a follow-up.

## Measurements

Local interleaved A/B on an M4 (`Apple M4, 10 vCPU`), 3 rounds, against this PR's base:

| Benchmark | base (#259) | this PR | |
| --- | --- | --- | --- |
| `parse_message` | 0.156 ms | 0.156 ms | +0.1% |
| `parse_metadata` | 0.029 ms | 0.029 ms | +0.7% |
| `parse_lazy_untouched` | 0.030 ms | 0.030 ms | +0.3% |
| `parse_lazy_all_attachments` | 0.163 ms | 0.164 ms | +0.8% |
| `parse_many` | 1.289 ms | 1.300 ms | +0.8% |
| `parse_many_metadata` | 0.229 ms | 0.229 ms | +0.1% |
| `parse_tree` | 0.156 ms | 0.156 ms | −0.3% |
| `parse_tree_metadata` | 0.030 ms | 0.030 ms | +0.0% |
| `parse_tree_lazy_untouched` | 0.030 ms | 0.030 ms | +0.7% |

Noise floor from the pure-Python controls **1.4%**; worst treatment delta **+1.8%** (`parse_qp_message`); verdict *no significant difference*. The issue predicted "every one within about ±1%, nothing expected to improve" — that is what happened.

**Verification**: 864 Python tests; 13 core Rust tests, two of them new — the RFC 2183 rule and the identity derivation now have direct tests instead of only three end-to-end ones; `cargo clippy --workspace --all-targets -D warnings -W clippy::cast_possible_truncation`; `cargo fmt`; `mypy --strict`; `ruff`; `check_vendored_mailparse.sh`; `check_bench_lockfile.py`. 179K `cargo fuzz run -a parse_agreement` executions, zero crashes. `__init__.pyi`, `tests/`, `docs/`, `vendor/` and `src/` are untouched — the diff is the core and the changelog.


kurok:
Green on the first run: 11/11, benchmark gate `no significant difference` (worst +4.6%, threshold 7%, AMD EPYC 9V74).

The four checks that did not register are CodeQL's, for the same reason as on #259 — this repo uses GitHub's default setup for code scanning, which only runs on pull requests targeting the default branch. They will run once this retargets to master as the stack merges.

PR #262, opened 2026-09-17
Split the binding into modules, and share its duplicated getters and batch loops (#233)

### LINKED ISSUE #233: Split src/fast_mail_parser.rs into modules and dedupe the copied getters and parse_many result loops
## Problem

The crate root is the 1797-line PyO3 binding file (`Cargo.toml:32-35` -> `path = "src/fast_mail_parser.rs"`), and `mod mail_parser;` at `src/fast_mail_parser.rs:17` is the only module boundary in the crate. Everything else -- four exceptions and the error mapping, the panic backstop, eleven `#[pyclass]`es across three modes and the tree API, three `#[pyfunction]`s with three mode-dispatch helpers each, payload handling and the `#[pymodule]` -- shares one file. The PyO3-free core (`src/mail_parser.rs`, 1463 lines) is a second single-file unit.

Inside the binding the same bodies are pasted rather than shared: the `headers` getter appears six times, `date_parsed` three times, `children` three times, the lazy `decode` twice, the `parse_many` result-collection loop three times and the `strict` gate four times. The file already states the rule this violates, on `resolve_workers` (`src/fast_mail_parser.rs:1514-1517`): *"Out of line and shared by all three modes rather than written three times: the message is part of the API, and three copies of it are three chances for them to stop agreeing."* The `strict` rejection and the `raise_on_error` slot semantics are part of the API in exactly that sense, and today a fix to one copy has to be replayed by hand in the other two or three.

This is a maintainability change, not a speed change. None of the duplicated code is on the parse path: the getters run on Python attribute access, the batch loops run with the GIL re-acquired after the detached parallel parse has finished, and the dominant costs of a full parse are the base64 decode (53%), the whitespace strip (27%) and `memmem` (7.5%), all inside the vendored mailparse. The library's reason to exist is speed, which is precisely why the binding around the hot core should be small enough that a change to one mode can be reviewed without loading the other 1700 lines -- and why every `#[inline(never)]`/`#[cold]`/`#[inline(always)]` decision in this file must survive the move untouched.

## Evidence

- `Cargo.toml:32-35` -- `[lib] name = "fast_mail_parser" path = "src/fast_mail_parser.rs" crate-type = ["cdylib"]`: the binding file is the crate root. `src/fast_mail_parser.rs:17` -- `mod mail_parser;` is the sole module declaration.
- Six byte-identical `headers` getters: `src/fast_mail_parser.rs:265-271` (`PyMailMetadata`), `360-366` (`PyMimePart`), `457-463` (`PyMimePartMetadata`), `525-531` (`PyLazyMimePart`), `863-869` (`PyMail`), `1105-1111` (`PyLazyMail`). Each is:
  ```rust
  fn headers<'py>(&self, py: Python<'py>) -> PyResult<Bound<'py, PyDict>> {
      let dict = PyDict::new(py);
      for (name, values) in &self.headers {
          dict.set_item(name, values)?;
      }
      Ok(dict)
  }
  ```
  over the same field type `pub headers: Vec<(String, Vec<String>)>` (lines 258, 338, 435, 497, 826, 1092).
- Three identical `date_parsed` getters: `278-284`, `872-878`, `1115-1121`:
  ```rust
  let Some(epoch) = mail_parser::parse_date_epoch(&self.date) else { return Ok(None); };
  let utc = PyTzInfo::utc(py)?;
  PyDateTime::from_timestamp(py, epoch as f64, Some(&utc)).map(Some)
  ```
- Three identical `children` getters: `381-383`, `467-469`, `535-537` -- `PyList::new(py, self.children.iter().map(|child| child.clone_ref(py)))`.
- Two identical decode-and-cache bodies: `PyLazyMimePart::decode` `589-596` and `PyLazyAttachment::decode` `1047-1055` -- `py.detach(|| mail_parser::decode_part(raw)).map_err(to_py_err)?; let bytes = PyBytes::new(py, decoded.as_slice()).unbind(); Ok(self.content.get_or_init(|| bytes).bind(py).clone())`. Both carry `#[cold] #[inline(never)]` (587-588, 1045-1046).
- Three copies of the `parse_many` result loop: `1550-1577` (full), `1609-1624` (metadata), `1656-1681` (lazy). The shared tail in every copy is `if raise_on_error { return Err(err); } items.append(err.value(py))?;`. The metadata copy has no strict arm, so it is structurally different from the other two even though it means the same thing.
- Four copies of the strict gate `if strict && !mail.warnings.is_empty() { Err(strict_rejection(&mail.warnings)) }`: `1335-1337`, `1341-1343`, `1558-1562`, `1663-1667`.
- `src/fast_mail_parser.rs:1514-1517` -- the maintainers' own rule quoted above, applied to `resolve_workers` only.
- Attributes that must survive verbatim: 25 `#[inline(never)]`, 9 `#[cold]`, and one `#[inline(always)]` on `catch_panics` (`174-175`), whose doc at `169-173` records the #99 incident: once a third instantiation existed the `parse_email` wrapper stopped inlining and cost 24% on large messages.
- Layout sensitivity (why the gate may fire on a pure move): `CHANGELOG.md:27-29` -- *"The version bump itself moved the decoding paths +3-7% against the last master commit on the M4"*; `CONTRIBUTING.md:330-334` -- symbol names and link order decide where the remaining `data_encoding` loop lands on the runners' Zen CPUs. Known profile: full parse 0.225 ms (M4) / 0.604 ms (EPYC), metadata 0.030 ms, `parse_many(8, threads=1)` 1.84 ms.

## Proposal

Two phases, both binding-only, shipped together in one PR. An optional third phase for the core is described separately and must be its own PR.

**Phase 1 -- shared helpers (the dedupe).** Add `src/convert.rs` with:

```rust
pub(crate) fn headers_dict<'py>(py: Python<'py>, headers: &[(String, Vec<String>)]) -> PyResult<Bound<'py, PyDict>>;
pub(crate) fn date_parsed<'py>(py: Python<'py>, date: &str) -> PyResult<Option<Bound<'py, PyDateTime>>>;
pub(crate) fn children_list<'py, T: pyo3::PyClass>(py: Python<'py>, children: &[Py<T>]) -> PyResult<Bound<'py, PyList>>;
#[cold] #[inline(never)]
pub(crate) fn decode_into<'py>(py: Python<'py>, cell: &OnceLock<Py<PyBytes>>, raw: &[u8]) -> PyResult<Bound<'py, PyBytes>>;
pub(crate) fn strict_gate(strict: bool, warnings: &[ParseWarning]) -> PyResult<()>;
/// One notion of "this slot failed", two ways to reach it (parse error or strict rejection).
pub(crate) fn push_slot<'py>(items: &Bound<'py, PyList>, outcome: PyResult<Py<PyAny>>, raise_on_error: bool) -> PyResult<()>;
```

Every getter listed under Evidence becomes a one-line delegation; the two `decode` methods become calls to `decode_into` (the `#[cold] #[inline(never)]` moves onto `decode_into`, and the two thin wrappers keep their attributes so the cached fast path in the getter above them is unchanged); the three `parse_many_*` loops become `for result in parsed { push_slot(&items, outcome(result), raise_on_error)?; }` with a per-mode `outcome` closure. Behaviour that must stay identical: header dict key order (wire order, #157), `date_parsed` returning an aware UTC `datetime` or `None`, the exception object (not a raise) appended to the slot when `raise_on_error=False`, the exact `strict_rejection` message, and `ParseError` subclass chosen by `to_py_err`.

**Phase 2 -- module split of the binding.** Keep `Cargo.toml` `[lib] path = "src/fast_mail_parser.rs"` and the filename unchanged (see traps), and reduce that file to the module docs, `mod` declarations, and the `#[pymodule]` (currently `1774-1797`). New files, all directly under `src/`:

| file | moves from `src/fast_mail_parser.rs` |
|---|---|
| `errors.rs` | the four `create_exception!`s, `to_py_err` (67), `strict_rejection` (103), `panic_message` (142), `catch_panics` (175, keep `#[inline(always)]`), `unknown_mode` (1350), `strict_needs_decoded_bodies` (1358) |
| `payload.rs` | `Payload` + `AsRef` impl (1188-1201), `payload_to_bytes` (1202), `resolve_workers` (1519) |
| `convert.rs` | Phase 1 helpers + `addresses` (299) |
| `metadata.rs` | `PyAttachmentMetadata`, `PyMailMetadata` (196-322) |
| `tree.rs` | `PyMimePart`, `PyMimePartMetadata`, `PyLazyMimePart`, `metadata_node`, `lazy_node` (332-664) |
| `flat.rs` | `ParseWarning`, `PyAddress`, `PyAttachment`, `PyMail` (671-911) |
| `lazy.rs` | `PyLazyAttachment`, `PyLazyMail` (925-1163) |
| `api.rs` | `parse_email`, `parse_many`, `parse_email_tree`, their `*_other_mode` / `*_strict_mode` / `*_inner` / `parse_many_*` helpers, `_panic_for_tests`, `panic_now` (1236-1772) |

Every `#[inline(never)]`, `#[cold]` and `#[inline(always)]` stays on the item it is on today; module boundaries do not change inlining under `lto = true, codegen-units = 1` (`Cargo.toml` release profile). Visibility widens from private to `pub(crate)` only where a moved item is used across files. No new dependency.

**Phase 3 (optional, separate PR, only if Phases 1-2 land with a flat gate).** Split `src/mail_parser.rs` into `src/mail_parser/{mod.rs, warnings.rs, repair.rs, batch.rs, headers.rs, flat.rs, tree.rs}`. This touches four external couplings (listed under traps) and its benefit is smaller -- the core is already a coherent PyO3-free unit whose module doc (`src/mail_parser.rs:1-37`) carries the load-bearing thread-safety invariant -- so it is not required to close this issue.

1. Read `src/fast_mail_parser.rs` in full once, and `CONTRIBUTING.md` sections "Performance" (about line 285-366) and the fuzz notes (185-215), before moving anything.
2. Phase 1 first, as its own commit: add `convert.rs`, replace the six/three/three/two/three/four copies, run `pytest tests --ignore=tests/benchmark`. The oracle for the getters is `tests/test_contract.py` (frozen export/attribute sets), `tests/test_stub_matches_runtime.py`, the `#157` header-order tests, and `tests/test_stdlib_parity.py` (respect its `DIVERGENCES`). The oracle for the batch loops is every test that calls `parse_many` with `raise_on_error=False` / `strict=True` across all three modes (grep `tests/` for `raise_on_error` and `strict=True`).
3. Phase 2 as a second commit: move code without editing it. Keep doc comments attached to their items (several are load-bearing: the `catch_panics` #99 note, the `resolve_workers` note, the lazy `decode` race note). Update the module-level `//!` docs at the top of `src/fast_mail_parser.rs` so they still describe the layout truthfully.
4. `cargo fmt --all -- --check` and `cargo clippy --all-targets -- -D warnings -W clippy::cast_possible_truncation` re-lint every moved line; expect `dead_code`/visibility warnings to surface and fix them by narrowing to `pub(crate)`, never by `#[allow]`.
5. Traps:
   - **sdist include is one level deep.** `pyproject.toml:46-48` -- `{ path = "src/*", format = "sdist" }`. New binding files must live directly under `src/`, not in `src/binding/`, or the sdist ships without them. The CI sdist smoke (`.github/workflows/test.yml:533-556`) builds from the sdist and would catch it, but only after the fact.
   - **CI hard-codes the two file names.** `.github/workflows/test.yml:542-543` asserts `src/mail_parser.rs` and `src/fast_mail_parser.rs` are in the sdist. Keeping both filenames avoids editing that list; if you rename the root to `lib.rs` anyway, update `Cargo.toml [lib] path` and that list together.
   - **`catch_panics` must keep `#[inline(always)]`** (`174`) and `children_list<T>` must stay a tiny non-recursive generic; the #99 incident was a generic wrapping the whole parse body, and `decode_into` must keep `#[cold] #[inline(never)]` so the hot cached path in the lazy getters is not disturbed.
   - **Phase 3 only:** `fuzz/fuzz_targets/parse_email.rs:22` and `parse_agreement.rs:48` include the core with `#[path = "../../src/mail_parser.rs"]` (rationale `fuzz/Cargo.toml:1-9`, `CONTRIBUTING.md:195-198`, `test.yml:182-184`); rustc treats a `#[path]`-loaded file as owning its directory, so `#[path = "../../src/mail_parser/mod.rs"]` should resolve `mod warnings;` etc. -- verify with `cargo +nightly fuzz build` before relying on it. `tests/test_readme_snippets.py:25,111` regex-scans `src/mail_parser.rs` for `KIND_[A-Z_]+: &str = "..."` and asserts the set is non-empty; point `SOURCE` at the new `warnings.rs`. `pyproject.toml:47` must become `src/**/*` for a subdirectory to reach the sdist, and `test.yml:542` must name a file that still exists. Keep the thread-safety invariant doc in `mod.rs`.
   - `vendor/mailparse` is not touched by this issue; do not run cargo inside it without `--target-dir` elsewhere (the sdist glob sweeps `target/` in).
   - Python surface is unchanged: `__init__.pyi`, `tests/test_contract.py` frozen sets and `docs/compatibility.md` need no edits; if any test there fails, the move changed behaviour and the move is wrong.
6. CHANGELOG: one line under Unreleased / Changed, "Internal: binding layer split into modules; shared getters and batch-result helpers. No behaviour or performance change intended."

## Harness -- how to prove it

Correctness and lint, all must pass locally:

```sh
cargo fmt --all -- --check
ruff check .
# Phase 3 only:
cargo +nightly fuzz build --manifest-path fuzz/Cargo.toml
# only if vendor/ is touched (it should not be):
```

Performance, interleaved A/B exactly as the CI gate does it (`.github/workflows/test.yml:444-468`): build master `73f56da` into `venv-base` and the branch into `venv-head` with `maturin develop --release` in each, then

```sh
for round in 1 2 3; do
  for side in head base; do
    "$side-venv/bin/pytest" -q tests/benchmark --benchmark-min-rounds=25 --benchmark-json="$side-$round.json"
  done
done
  --a-label "this revision" --b-label base \
  --informational test__threaded___ \
```

Benchmarks that exercise the moved code and are **expected flat (within the noise floor set by the pure-Python controls)**: `test__fast_mail_parser___full_read` (reads `headers`, `date_parsed`, attachments -- the deduped getters), `test__fast_mail_parser___parse_many`, `test__fast_mail_parser___parse_many_metadata` (the shared `push_slot` loop), `test__fast_mail_parser___parse_lazy_all_attachments` (`decode_into`), plus `test__fast_mail_parser___parse_message`, `___parse_message_strict`, `___parse_metadata`, `___parse_tree`. No benchmark is expected to improve; no benchmark may regress more than 7%. No new benchmark is needed.

If the gate reports a decode-path regression (`parse_message`, `full_read`, `parse_many`, `parse_lazy_all_attachments`) while the metadata paths and the pure-Python controls are flat, that is the code-layout artifact `CHANGELOG.md:27-29` and `CONTRIBUTING.md:330-334` describe -- symbol names and link order changed, the `data_encoding` loop moved. Re-run the three rounds before acting; a real regression reproduces, a layout artifact does not. The M4 is not a proxy for x86: confirm on the PR's CI gate (AMD EPYC), and if it is still ambiguous, `gh workflow run toolchain-ab.yml` against the branch. Put the local before/after `ab_median.py` table and the CPU it ran on in the PR body.

One extra correctness check for Phase 1, since the three batch loops were not identical: add a test (e.g. `tests/test_parse_many_modes.py::test__parse_many___slot_semantics_agree_across_modes`) that feeds the same batch of `[good, malformed, lossy]` payloads to `parse_many(mode=m, raise_on_error=False, strict=s)` for every `(m, s)` in `{full, lazy} x {False, True}` plus `(metadata, False)`, and asserts that the slot is a `ParseError` instance exactly where the mode/strict combination says it must be, and that `raise_on_error=True` raises the same exception type the slot would have held. This pins the semantics the shared helper now owns.

## Acceptance criteria

- [ ] `src/fast_mail_parser.rs` contains only module docs, `mod` declarations and the `#[pymodule]`; the pyclasses, pyfunctions, errors, payload handling and converters live in `src/{errors,payload,convert,metadata,tree,flat,lazy,api}.rs` directly under `src/`.
- [ ] Exactly one definition each of the headers-dict, date-parsed, children-list, decode-into-cache, strict-gate and slot-append logic; `grep -c 'dict.set_item(name, values)' src/*.rs` returns 1.
- [ ] Every `#[inline(never)]`, `#[cold]` and `#[inline(always)]` present at `73f56da` is present on the same item after the move (`grep -c` counts 25 / 9 / 1 across `src/*.rs`, excluding `src/mail_parser.rs`).
- [ ] No edits to `fast_mail_parser/__init__.py`, `fast_mail_parser/__init__.pyi`, `tests/test_contract.py`, `docs/compatibility.md`, or `vendor/`.
- [ ] `pytest tests --ignore=tests/benchmark`, `cargo fmt --all -- --check`, `cargo clippy --all-targets -- -D warnings -W clippy::cast_possible_truncation`, `mypy --strict fast_mail_parser/`, `ruff check .` all pass.
- [ ] The sdist smoke job passes (new files are picked up by `src/*`).
- [ ] Interleaved 3-round `ab_median.py` against `73f56da`: no benchmark regresses more than 7%, the pure-Python controls are within their usual floor, and the table is in the PR body with the CPU named.
- [ ] New `test__parse_many___slot_semantics_agree_across_modes` (or equivalent) is present and passes.
- [ ] CHANGELOG has an Unreleased entry stating no behaviour or performance change is intended.
- [ ] (Phase 3 only, separate PR) `cargo +nightly fuzz build` succeeds with the new `#[path]`, `tests/test_readme_snippets.py` still finds the `KIND_*` constants, `pyproject.toml` include reaches the subdirectory, and `test.yml:542-543` names files that exist.

## Out of scope

- Any change to parsing behaviour, warning kinds, error messages or the Python surface.
- Any change under `vendor/mailparse` or to the release profile / toolchain pin.
- Performance work on the remaining hot loops (`data_encoding` base64 decode, charset conversion).
- Renaming the `#[pyclass]` types or setting `#[pyclass(module = ...)]`.
- Turning the fuzz harness into a workspace member or making the core a separate crate.

## Related issues

- #231 — the `headers` getter collapse is specified there (with caching); land it first, then this issue's Phase 1 covers the remaining helpers
- #234 — the core-side dedupe
- #236 — the crate-layout half; this issue is binding-only

kurok:
Closes #233 (phases 1 and 2; phase 3 is out of scope in the issue itself).

> **Stacked on #261** → #260 → #259 → #258. Each retargets itself as the one below merges. Two commits: the dedupe, then the move.

## Phase 1 — the dedupe

`date_parsed` had three byte-identical copies, `children` three, the decode-and-cache body two, the strict gate four, and the `parse_many` result loop three. They are now `src/convert.rs`.

The file already stated the rule, on `resolve_workers`: *"Out of line and shared by all three modes rather than written three times: the message is part of the API, and three copies of it are three chances for them to stop agreeing."* The strict rejection's text and what `raise_on_error=False` puts in a failed slot are part of the API in exactly that sense.

**One item on the issue's list is not here: `headers_dict`.** The issue cites six byte-identical five-line `headers` getters — but that evidence predates **#231**, which replaced them with a cached `Headers::to_dict`. All six are already one-line delegations, and the dict is memoised as well as shared, which is strictly better than what the issue asked for. Nothing to do.

`decode_into` keeps the `#[cold] #[inline(never)]` both copies carried, so the cached fast path in the getters above it is undisturbed, and `children_list` is a tiny non-recursive generic — the #99 incident was a generic wrapping the whole parse body, which is not this shape. Behaviour is unchanged including *which* errors propagate out of a batch rather than landing in a slot: `Py::new`'s fallibility stays outside the outcome, exactly where it was.

## Phase 2 — the split

| module | what is in it |
|---|---|
| `errors.rs` | the four exception types, `to_py_err`, `strict_rejection`, the panic backstop |
| `payload.rs` | `Payload`, `payload_to_bytes`, `Pinned`, `RetainedBytes` |
| `convert.rs` | `Headers`, `addresses`, phase 1's helpers |
| `metadata.rs` | `PyAttachmentMetadata`, `PyMailMetadata` |
| `flat.rs` | `ParseWarning`, `PyAddress`, `PyAttachment`, `PyMail` |
| `lazy.rs` | `PyLazyAttachment`, `PyLazyMail` |
| `tree.rs` | the three node types, `metadata_node`, `lazy_node` |
| `api.rs` | the three `#[pyfunction]`s and their mode dispatch |

2000 lines → an 85-line root that is the module list and the `#[pymodule]`. All eight files sit directly under `src/`, because the sdist include glob is `src/*` and is one level deep.

Code was **moved, not edited**. Visibility widened from private to `pub(crate)` only where a moved item is read across files, never with `#[allow(dead_code)]`.

## The checks that matter for a pure move

**Attributes survived — after a correction.** Final counts are `#[inline(never)]` 23, `#[cold]` 7, `#[inline(always)]` 1, against a pre-#233 baseline of 22, 6 and 1. The `+1` on each is `decode_into` itself, which is what the issue asked for: the attributes move onto the shared helper **and** the two thin wrappers keep theirs.

I got that wrong first time round. Phase 1 dropped `#[cold] #[inline(never)]` from both `decode` wrappers, and my original check compared the commit before the split against the commit after — which verified the *move* changed nothing but silently accepted phase 1's loss. The third commit restores them. See the comment below for how the gate caught it; the lesson is that the baseline for an attribute count is the commit before the whole PR, not before its last step.

**The sdist really carries them.** Built locally with `python -m build --sdist` and listed: all eight modules present. CI's sdist manifest check now *names* them rather than trusting the glob, so a glob that stopped matching fails by naming the missing file instead of surfacing as a confusing compile error later in that job.

**Nothing public moved.** `__init__.pyi`, `tests/test_contract.py`'s frozen sets, `docs/` and `vendor/` are untouched; 867 Python tests pass, including `test_contract.py`, `test_stub_matches_runtime.py`, the #157 header-order tests and `test_stdlib_parity.py` with no `DIVERGENCES` edit.

## Measurements

Local interleaved A/B on an M4, 3 rounds, against this PR's base:

| Benchmark | base (#261) | this PR | |
| --- | --- | --- | --- |
| `parse_message` | 0.158 ms | 0.156 ms | −1.0% |
| `parse_metadata` | 0.029 ms | 0.029 ms | +0.1% |
| `parse_tree` | 0.155 ms | 0.155 ms | −0.1% |
| `parse_lazy_untouched` | 0.031 ms | 0.030 ms | −0.4% |
| `parse_many` | 1.310 ms | 1.290 ms | −1.5% |
| `parse_qp_message` | 0.130 ms | 0.131 ms | **+1.2%** (worst) |

Noise floor **2.2%**; worst real movement **+1.2%**. `ab_median.py` reports `attachment_reread` at "+97.6%" in this PR's favour — ignore it, as on #259: that benchmark is 42–83 ns and both sides sample the same two timer ticks, so the median flips between them. It is quantisation, not a measurement.

**Verification**: 867 Python tests; `cargo clippy --workspace --all-targets -D warnings -W clippy::cast_possible_truncation`; `cargo fmt`; `mypy --strict`; `ruff`; sdist built and inspected.


kurok:
### The gate caught a real regression, and it was mine

The benchmark gate failed twice on `parse_qp_message`, +8.2% then +8.3%, on `AMD EPYC 7763` both times — base 0.247/0.248 ms against 0.267/0.268 ms. Reproducible, not a noisy draw.

**My first instinct was wrong and I want to record it, because it is the instinct this repo's history encourages.** This PR is binding-only: `git diff base..HEAD -- crates/ vendor/` is **0 bytes**, so the quoted-printable decoder is byte-identical. The obvious reading was #204/#240 code-layout scatter, and `parse_qp_message` is the benchmark with exactly that history — three earlier PRs failed on it at +9.2%, +7.3% and +19.0% for changes that could not reach QP decoding.

So I measured it instead of asserting it, with the instrument #240 built for this. `layout-ab.yml`, 4 salted builds of this branch, 3 interleaved rounds:

| | spread across salts |
| --- | --- |
| `parse_qp_message` | **+1.1%** |
| `parse_message` | +1.6% |
| `parse_qp_dense_escapes` | +0.4% |
| control floor (`mailparser_lib___full_read`) | +6.6% |

`parse_qp_message` is **not** link-order sensitive on this codebase — 1.1% across four differently-salted builds. An 8.3% delta could not be explained away as layout, and the hypothesis was dead.

**The actual cause.** The issue specifies that the `#[cold] #[inline(never)]` moves onto `decode_into` *and* "the two thin wrappers keep their attributes so the cached fast path in the getter above them is unchanged". Phase 1 dropped them from both wrappers. Their own doc comments still read "`#[cold]` and out of line", which is what gave it away.

My attribute check had missed it because I compared the commit **before the split** against the commit **after** — which correctly proved the move changed nothing, and silently inherited phase 1's loss. Against the real baseline the counts were 21/5 where they should have been 23/7.

**After restoring them, on the same CPU class as both failures:**

| `parse_qp_message` | base | this PR | |
| --- | --- | --- | --- |
| EPYC 7763, before the fix (×2) | 0.247 / 0.248 ms | 0.267 / 0.268 ms | +8.2%, +8.3% |
| EPYC 7763, after the fix | 0.246 ms | 0.252 ms | **+2.3%** |
| Xeon 8573C, after the fix | 0.259 ms | 0.264 ms | +1.8% |

Same runner class, same base value, treatment down from 0.267 to 0.252. That is causal, not a luckier draw — the first passing run was on a different CPU and I did not want to rest on it, so I re-ran until one landed on 7763.

Worst treatment delta is now +3.5% (`parse_qp_dense_escapes`) against a 1.0% noise floor, verdict *no significant difference*, 11/11 green.

Two things worth keeping from this: the gate earned its keep here — an organisational refactor was quietly costing 8% on a decode path — and "it's layout" is a hypothesis that this repo now has a tool to test rather than a conclusion to reach for.

